PR #19, opened 2016-07-21
Qc issue2

### LINKED ISSUE #2: Save Image Link
#### As a user I want to save a link to an image.

Acceptance criteria:
- [ ] The link is entered through a form.
- [ ] After the form is saved, the link is persisted in the database.
- [ ] After the form is submitted, the user is redirected to a page displaying the image.

Dependencies:
- Deployment Environment

__Note__: Many image sharing sites will permit one to upload images, or
download images when provided a link. For this project we will only work with
the link, i.e., URL, to various images.

mqchau:
Closes #2


bboe:
Can you assign this to @leoi11? He's the assignee.

bboe:
@mqchau it's awesome that you used the scaffolding. Now, unfortunately, I'm going to ask you to remove most of it.

While scaffolding makes developing applications in rails very convenient, it makes actually completing stories much more cumbersome. All of those actions that you added require extensive tests to go along with them which scaffolding doesn't provide. As a result we could spend the next week iterating on this one story in order to get everything done that's necessary. That's too much work for a single pull request.

Thus, please only implement the new and create actions for this pull request. In general, ensure that every file you add to a pull request serves a purpose. Just because rails adds it for you doesn't mean it's needed at that time -- you can always add it when you do need it.

Also please add all the unit and controller tests that you think are necessary to test these features you've added. Every single feature addition commit should include its necessary tests. I think it's a good idea to break up the new action into its own commit, and the create action into its own commit. Since neither Leo nor I have actually commented on anything in this PR please fixup your history prior to the initial review. Thanks!

mqchau:
Ok I'll work on it.

bboe:
I didn't realize it (it's a newer addition) but the show action should also be included in this story. My mistake.

app/assets/javascripts/image_links.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
```
bboe: What purpose does this file provide to this feature?

app/assets/stylesheets/image_links.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the ImageLinks controller here.
```
bboe: What purpose does this file provide to this feature?

app/assets/stylesheets/scaffolds.scss
```diff
@@ -0,0 +1,89 @@
+body {
```
bboe: What purpose does this file provide to this feature?

app/controllers/image_links_controller.rb
```diff
@@ -0,0 +1,41 @@
+class ImageLinksController < ApplicationController
+  before_action :set_image_link, only: [:show, :edit, :update, :destroy]
+
+  # GET /image_links/1
```
bboe: Can we remove these scaffold generated comments? They're great for someone new to rails, but we want to avoid these sort of comments in our code.

app/assets/stylesheets/image_links.scss
mqchau: This file maybe useful when it comes to issue #6 to add style. Can we leave it as place holder here?

app/controllers/image_links_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  def new
+    @image_link = ImageLink.new
+  end
+
+  # POST /image_links
+  # POST /image_links.json
+  def create
+    @image_link = ImageLink.new(image_link_params)
+
+    respond_to do |format|
```
bboe: What purpose does the json responses serve as part of this story?

app/controllers/image_links_controller.rb
```diff
@@ -0,0 +1,41 @@
+class ImageLinksController < ApplicationController
+  before_action :set_image_link, only: [:show, :edit, :update, :destroy]
```
bboe: Do we have an `edit`, `update`, `destroy`?

app/assets/stylesheets/scaffolds.scss
mqchau: This file is useful when it comes to issue #6 to add style. Can we leave it here?

app/controllers/image_links_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  # POST /image_links.json
+  def create
+    @image_link = ImageLink.new(image_link_params)
+
+    respond_to do |format|
+      if @image_link.save
+        format.html { redirect_to @image_link, notice: 'Image link was successfully created.' }
+        format.json { render :show, status: :created, location: @image_link }
+      else
+        format.html { render :new }
```
bboe: How do we indicate that this HTTP request resulted in a failure?

app/helpers/image_links_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ImageLinksHelper
```
bboe: What purpose does this file provide to this feature?

app/views/image_links/show.html.erb
```diff
@@ -0,0 +1,9 @@
+<p id="notice"><%= notice %></p>
+
+<p>
+  <strong>Link:</strong>
+  <%= @image_link.link %>
+  </br>
+  <img src="<%= @image_link.link %>"></img>
```
bboe: Is there a rails way to render this image tag?

app/views/image_links/show.html.erb
```diff
@@ -0,0 +1,9 @@
+<p id="notice"><%= notice %></p>
+
+<p>
+  <strong>Link:</strong>
```
bboe: Do you think outputting the URL is necessary?

config/routes.rb
```diff
@@ -1,4 +1,9 @@
 Rails.application.routes.draw do
   root 'welcome#index'
+
+  # resources :image_links
+  get '/image_links/new', to: 'image_links#new', as: 'new_image_link'
```
bboe: Is there a more concise way to handle these three statements? (Hint: a part of it is in a comment).

db/seeds.rb
```diff
@@ -5,3 +5,6 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+#
+
+ImageLink.create([{ link: 'http://i.imgur.com/PVrX7wZ.jpg' }])
```
bboe: What will happen if this create fails? How can we be sure that such a failure isn't silently dropped?

leoi11:
Consider Image as a RESTful resource, and url/link is one of the attribute of an image object.
The Rails convention is to name the controllers, models according to the resource name. In this case, it would be more appropriate to call this resource Image instead of ImageLink because in the future this image link object will actually have other attributes like name, source, etc. And these attributes do not really belong to a ImageLink object, and wouldn't make sense to belong to it.

test/controllers/image_links_controller_test.rb
```diff
@@ -0,0 +1,29 @@
+require 'test_helper'
+
+class ImageLinksControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    @image_link = image_links(:one)
```
bboe: It's a good practice to avoid the use of fixtures. While the presence of the fixtures file is useful to ensure the table is empty on each test run, the actual use of fixtures is discouraged. We prefer to actually create the objects we require in the test case itself -- or in a helper function that the test explicitly calls.

test/controllers/image_links_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  test 'should get new' do
+    get new_image_link_url
+    assert_response :success
+    assert_select 'form'
```
bboe: Can we be more specific than just `form`? We might have multiple form templates. What element on the form indicates that it's the form for creating a ImageLink?

app/controllers/image_links_controller.rb
mqchau: So do you want to return HTTP status code 400 or something similar? Right now if this request fails it will return HTTP status 200 and the /new page with error notes on it.

app/views/image_links/show.html.erb
mqchau: Fixed it

test/controllers/image_links_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  test 'should create image_link' do
+    assert_difference('ImageLink.count', 1) do
+      post image_links_url, params: { image_link: { link: @image_link.link } }
+    end
+
+    assert_redirected_to image_link_url(ImageLink.last)
```
bboe: `image_link_url(ImageLink.last)` I think this can just be shortened to `ImageLink.last`

test/controllers/image_links_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  test 'should get new' do
+    get new_image_link_url
+    assert_response :success
+    assert_select 'form'
+  end
+
+  test 'should create image_link' do
+    assert_difference('ImageLink.count', 1) do
```
bboe: ⭐ Great job using `assert_difference`

test/controllers/image_links_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    @image_link = image_links(:one)
+  end
+
+  test 'should get new' do
+    get new_image_link_url
+    assert_response :success
+    assert_select 'form'
+  end
+
+  test 'should create image_link' do
```
bboe: Should we write a failure test case?

test/models/image_link_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class ImageLinkTest < ActiveSupport::TestCase
+  # test "the truth" do
```
bboe: What do we need to test here?

test/controllers/image_links_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+      post image_links_url, params: { image_link: { link: @image_link.link } }
+    end
+
+    assert_redirected_to image_link_url(ImageLink.last)
+  end
+
+  test 'should show image_link and display the image' do
+    get image_link_url(@image_link)
+    assert_response :success
+    assert_select 'img' do
```
bboe: ⭐  Great test. I especially like that you look for the image link in the response.

test/controllers/image_links_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  test 'should create image_link' do
+    assert_difference('ImageLink.count', 1) do
+      post image_links_url, params: { image_link: { link: @image_link.link } }
+    end
+
+    assert_redirected_to image_link_url(ImageLink.last)
+  end
+
+  test 'should show image_link and display the image' do
```
bboe: What happens if we try to get an image by an ID that doesn't exist?

mqchau:
See pull request #23

PR #25, opened 2016-07-22
Qc check url

### LINKED ISSUE #4: Image URL Validation
#### As a user I want to be prevented from submitting invalid image URLs.

Acceptance criteria:
- [ ] I cannot successfully save an image with an invalid URL.
- [ ] An error message is associated with the appropriate input field on
  failure.

Discussion Topic:
- [ ] What is a valid image URL?

Dependencies:
- Save Image Link

mqchau:
Add URL validation in Image Model

Add errors styling in form

Closes #4, depends on #2


app/assets/stylesheets/scaffolds.scss
```diff
@@ -0,0 +1,89 @@
+body {
```
XanderStrike: This change seems a little out of place for this pull and especially this commit.

app/models/image.rb
```diff
@@ -1,2 +1,3 @@
 class Image < ApplicationRecord
+  validates :link, url: true
```
XanderStrike: This works great! It's up to you, but I usually recommend validations on type as well, meaning validating that the link ends with jpg or png or gif.

app/assets/stylesheets/scaffolds.scss
mqchau: Yeah I agree it would be more appropriate to put on another issue that deals with styling.

app/models/image.rb
mqchau: I was using httparty to make an actual request and check the response header content-type to start with "image/" but David Smith advice against it. I have seen some image links that don't end with jpg, png, gif... so I don't think we should check for that.

mqchau:
I removed scaffolds.scss as you said, Alex.

app/models/image.rb
XanderStrike: Yeah it's totally up to you.

test/controllers/images_controller_test.rb
```diff
@@ -21,6 +21,15 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'Image was successfully created.', flash[:notice]
   end
 
+  test 'should not create invalid image' do
+    assert_difference('Image.count', 0) do
+      post images_url, params: { image: { link: 'not_a_valid_url' } }
+      assert_response :unprocessable_entity
```
bboe: Do these three assert statements require being included within `assert_difference`?

test/models/image_test.rb
```diff
@@ -2,7 +2,22 @@
...
   test 'valid image' do
-    image = Image.new(link: 'sdfasfasf')
+    image = Image.new(link: 'http://www.abc.com/cde')
     assert_predicate image, :valid?
   end
+
+  test 'invalid image URL' do
+    image = Image.new(link: 'httpwww.abc.com/cde')
+    refute_predicate image, :valid?
+  end
```
bboe: Can you also assert that the error message is set on the image instance?

test/models/image_test.rb
```diff
@@ -2,7 +2,22 @@
...
+
+  test 'blank image URL should be invalid' do
+    image = Image.new(link: '')
+    refute_predicate image, :valid?
+  end
```
bboe: Can you also assert that the error message is set on the image instance?

test/models/image_test.rb
```diff
@@ -2,7 +2,22 @@
...
+
+  test 'nil image URL should be invalid' do
+    image = Image.new(link: nil)
+    refute_predicate image, :valid?
```
bboe: Can you also assert that the error message is set on the image instance?

bboe:
A few minor changes please.

mqchau:
Yeah I got you Bryce.

bboe:
Awesome. The very last thing, that I neglected earlier, is to modify your database such that an image cannot be inserted without a URL. The rails side is already doing this validation, but it's good practice to also have your database enforce the things that it can enforce. This change will require a migration.

mqchau:
Ok. Did the migration.

db/schema.rb
```diff
@@ -10,12 +10,37 @@
...
+ActiveRecord::Schema.define(version: 20160722224622) do
 
   create_table "images", force: :cascade do |t|
-    t.string   "link"
+    t.string   "link",       null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
   end
 
+  create_table "taggings", force: :cascade do |t|
```
bboe: I don't think you meant to include all of these changes.

db/migrate/20160722224622_require_image_link.rb
```diff
@@ -0,0 +1,5 @@
+class RequireImageLink < ActiveRecord::Migration[5.0]
+  def change
+    change_column_null :images, :link, false
```
bboe: ⭐

mqchau:
Yeah I don't know why that schema remains from another branch I was trying to do taggings

bboe:
The schema is generated from the state of the migrations table (if I'm not mistaken). To prevent that you either need to drop/recreate, or migrate down prior to switching branches.

bboe:
👍  Looks good to merge!

PR #27, opened 2016-07-25
add tags to images

### LINKED ISSUE #7: Image Tags
#### As a user I want to add tags to images.

Acceptance criteria:
- [ ] When I add a new image, there is a form field to add tags.
- [ ] After I've saved an image with tags, I see those tags displayed with the
  associated image on the index page.
- [ ] My application uses the acts-as-taggable-on gem to provide the tagging
  functionality.

Reference material:
- https://github.com/mbleigh/acts-as-taggable-on

Dependencies:
- Image Index

mqchau:
Closes #7


db/migrate/20160725164130_acts_as_taggable_on_migration.acts_as_taggable_on_engine.rb
```diff
@@ -0,0 +1,35 @@
...
+  def self.up
+    create_tag_table
+
+    add_index :taggings, :tag_id
+    add_index :taggings, [:taggable_id, :taggable_type, :context]
+  end
+
+  def create_tag_table
+    create_table :tags do |t|
+      t.string :name
```
shfung: Maybe we can add some not-null constraints to some of the columns?

mqchau: Yeah maybe we can make that constraint not null on table tags column name and table taggings column created_at

test/controllers/images_controller_test.rb
```diff
@@ -18,7 +20,7 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
 
   test 'should create image' do
     assert_difference('Image.count', 1) do
-      post images_url, params: { image: { link: @sample_img_link } }
+      post images_url, params: { image: { link: @sample_img_link, tag_list: @sample_img_tags } }
```
shfung: Maybe add a test that creates an image without any tag?

test/models/image_test.rb
```diff
@@ -24,10 +26,22 @@ class ImageTest < ActiveSupport::TestCase
     assert_includes image.errors.full_messages, @invalid_img_url_err_msg
   end
 
+  test 'add single tag to image' do
```
shfung: I think this test name is a little bit confusing because it is not 'adding' a tag to any image, it's creating an image with a single tag.

test/models/image_test.rb
```diff
@@ -24,10 +26,22 @@ class ImageTest < ActiveSupport::TestCase
...
+  end
+
   test 'nil image URL should be invalid' do
     image = Image.new(link: nil)
     refute_predicate image, :valid?
     assert_predicate image.errors, :any?
     assert_includes image.errors.full_messages, @invalid_img_url_err_msg
   end
+
+  test 'add many tags to image' do
```
shfung: Similar comment as above.

app/views/images/index.html.erb
```diff
@@ -1,6 +1,7 @@
 <h1>Images</h1>
 
 <% @images.each do |image| %>
-  <%= image_tag(image.link, :class => 'medium_img_display') %>
+  <%= image_tag(image.link, :class => 'image__medium_display') %>
+  Tags: <p class='image__tag'><%= image.tag_list %></p>
```
shfung: Can you use double quote for html string?

app/views/images/show.html.erb
```diff
@@ -1,4 +1,6 @@
 <p>
   <%= image_tag(@image.link) %>
+  <br/>
+  Tags: <p class='image__tag'><%= @image.tag_list %></p>
```
shfung: Same comment as above.

db/migrate/20160725164130_acts_as_taggable_on_migration.acts_as_taggable_on_engine.rb
shfung: That's a good start.  But I think there are more columns that can have not null constraints.
It's generally a good practice to add not-null constraints to all columns, except the ones that actually allow null.

db/schema.rb
```diff
@@ -10,11 +10,37 @@
...
     t.datetime "updated_at", null: false
   end
+
+  create_table "taggings", force: :cascade do |t|
+    t.integer  "tag_id",                    null: false
+    t.string   "taggable_type",             null: false
+    t.integer  "taggable_id",               null: false
+    t.string   "tagger_type"
+    t.integer  "tagger_id"
+    t.string   "context",       limit: 128
```
shfung: I think `context` can also have non-null constraint.

db/schema.rb
```diff
@@ -10,11 +10,37 @@
...
+    t.index ["taggable_id", "taggable_type", "tagger_id", "context"], name: "taggings_idy"
+    t.index ["taggable_id"], name: "index_taggings_on_taggable_id"
+    t.index ["taggable_type"], name: "index_taggings_on_taggable_type"
+    t.index ["tagger_id", "tagger_type"], name: "index_taggings_on_tagger_id_and_tagger_type"
+    t.index ["tagger_id"], name: "index_taggings_on_tagger_id"
+  end
+
+  create_table "tags", force: :cascade do |t|
+    t.string  "name",                       null: false
+    t.integer "taggings_count", default: 0
```
shfung: I think `taggings_count ` can have non-null constraint as well.

test/models/image_test.rb
```diff
@@ -3,14 +3,16 @@
 class ImageTest < ActiveSupport::TestCase
   setup do
     @invalid_img_url_err_msg = 'Link is not a valid URL'
+    @sample_img_link = 'http://i.imgur.com/PVrX7wZ.jpg'
+    @sample_img_tags = 'testtag1, testtag2'
```
shfung: `@ sample_img_tags` is only used in one test, I wonder if it should be in the `setup` or just in that test only?

test/controllers/images_controller_test.rb
```diff
@@ -3,22 +3,42 @@
...
+    end
+
+    assert_redirected_to Image.last
+    assert_equal 'Image was successfully created.', flash[:notice]
+  end
+
+  test 'should create image with no tags' do
+    assert_difference('Image.count', 1) do
+      post images_url, params: { image: { link: @sample_img_link, tag_list: '' } }
```
shfung: In the `'should not create invalid image'` test, I think you should also add the key for `tag_list`.

PR #28, opened 2016-07-26
add buttons to let user delete images on index and show page

### LINKED ISSUE #8: Delete Image
#### As a user I want to delete an image.

Acceptance criteria:
- [ ] The user is prompted for confirmation prior to deletion.
- [ ] Deleted images are removed from the homepage.

Dependencies:
- Image Index

mqchau:
Closes #8


app/controllers/images_controller.rb
```diff
@@ -21,6 +21,11 @@ def create
     end
   end
 
+  def destroy
+    @image.destroy
```
D-D-M: I notice you're calling `set_image` before the destroy action. But, in the spirit of fault tolerance, what if two people try to delete this image at the same time? Buckle up, because I'm about to tell you what.

So let's say Alice and Bob try to delete image `id: 4` at the same time. Alice's request is received a split second before Bob's, so she deletes the image, and everything is fine. Now it's Bob's turn for his request to be processed. Down on [line 32 here](https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/28/files#diff-a7ae6992fe889ec0c720925f4275f636R32), Bob will search for the image, but find nothing:

`@image = Image.find(4) # Exception: ActiveRecord::RecordNotFound`

Bob gets an ugly error page.

A nice way to avoid this is to use the `find_by` method instead of `find`, because `find_by` will return `nil` whenever it doesn't get any results (it won't throw an exception). Then, in your destroy action, you can ask if the image was successfully found:

```
def destroy
  if @image.present?
    @image.destroy
    flash[:notice] = 'Image was successfully destroyed.'
  else
    flash[:failure] = 'Image does not exist.'
  end
  
  redirect to images_url
end
```

👍

app/controllers/images_controller.rb
```diff
@@ -21,6 +21,11 @@ def create
...
+    redirect_to images_url, notice: 'Image was successfully destroyed.'
```
D-D-M: This is a minor comment, but if you successfully complete an action, you typically set the flash `:success` message instead. So that could either be:
```
redirect_to images_url, success: 'Image was successfully destroyed."

# OR

flash[:success] = 'Image was successfully destroyed.'
```

...depending on which approach you'd like to take.

app/controllers/images_controller.rb
bboe: @D-D-M there is still a race condition in what you've suggested between @image.present? and @image.destroy. What I think should be done is to check the return value of destroy to see if it succeeded if you really want to be precise.

However, delete failing is an exceptional case so I think it's acceptable to simply use `destroy!` and ignore the failure case. If you want to handle it, then it's important to write a test case for it which can be a bit tricky.

test/controllers/images_controller_test.rb
```diff
@@ -60,4 +60,26 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
     assert_select 'h1', 'Images'
   end
+
+  test 'should delete image and should not show that image' do
+    image = Image.create!(link: @sample_img_link)
+    assert_difference('Image.count', -1) do
+      delete image_url(image)
+    end
+
```
D-D-M: I'm not familiar with this particular DSL, but it couldn't hurt to set an assertion about the flash message you're expecting:
```
assert_equal 'Image was successfully deleted.', flash[:success]
```

test/controllers/images_controller_test.rb
```diff
@@ -60,4 +60,26 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    assert_redirected_to images_url
+    assert_select "img[src='#{image.link}']", false
+
+    assert_raises(ActiveRecord::RecordNotFound) do
+      get image_url(image)
+    end
```
D-D-M: I'm not familiar with this particular DSL, but it couldn't hurt to set an assertion about the flash message you're expecting:
```
assert_equal 'Image does not exist.', flash[:failure]
```
Even though your story only has to do with deleting, I think it's a good idea to handle this case in the `#show` action.

test/controllers/images_controller_test.rb
```diff
@@ -60,4 +60,26 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  end
+
+  test 'should not delete image with invalid id' do
+    assert_difference('Image.count', 0) do
+      assert_raises(ActiveRecord::RecordNotFound) do
```
D-D-M: If you make the changes above where I suggested you use `find_by`, this can change to just a simple assertion about the flash message.
```
assert_difference('Image.count', 0) do
  delete image_url(-1) # -1 is common for nonexistent ids
end

assert_equal 'Image does not exist.', flash[:failure]
assert_redirected_to images_url
```

test/flows/image_flow_test.rb
```diff
@@ -9,4 +9,14 @@ class ImageFlowTest < FlowTestCase
     assert_equal 'Image was successfully created.', find('#notice').text
     assert_equal Image.last.link, find('img')[:src]
   end
+
+  test 'add new image then delete' do
+    visit(new_image_path)
+    fill_in('image[link]', with: 'http://i.imgur.com/PVrX7wZ.jpg')
+    find('input[type="submit"]').click
```
D-D-M: Capybara lets you use this nifty [click_button](http://www.rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Actions#click_button-instance_method) method. So this line could be more easily written like this:
`click_button 'Create Image'` (or whatever the text of this button is)

test/flows/image_flow_test.rb
```diff
@@ -9,4 +9,14 @@ class ImageFlowTest < FlowTestCase
...
+    assert has_current_path?(image_path(Image.last))
+    find('a[data-method="delete"]').click
```
D-D-M: Capybara also has a [click_link](http://www.rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Actions#click_link-instance_method) method 😄

app/controllers/images_controller.rb
D-D-M: Ahhh, good point.

Quan, feel free to ignore the failure case :)

app/controllers/images_controller.rb
```diff
@@ -15,16 +15,24 @@ def new
...
+    flash[:success] = 'Image was successfully destroyed.'
+    redirect_to images_url
+  end
+
   private
 
   def set_image
-    @image = Image.find(params[:id])
+    @image = Image.find_by(id: params[:id])
+    render file: "#{Rails.root}/public/404", layout: false, status: :not_found if @image.nil?
```
bboe: This automatically happens when `find` fails on production. No need to explicitly handle it.

app/views/images/_flash_message.html.erb
```diff
@@ -0,0 +1,2 @@
+<p id="notice"><%= notice %></p>
```
bboe: How can you do this so you don't need to list all the possible flash message types? Also, how can you do it such that the `<p>` tags don't show up when there isn't a flash message?

app/views/layouts/application.html.erb
```diff
@@ -9,8 +9,8 @@
   </head>
 
   <body>
-    <p id="notice"><%= notice %></p>
 
+    <%= render partial: 'images/flash_message' %>
```
bboe: Why is the partial under the images scope?

test/controllers/images_controller_test.rb
```diff
@@ -63,9 +63,8 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
   end
 
   test 'should not show image with invalid id' do
-    assert_raises ActiveRecord::RecordNotFound do
-      get image_url(-1)
-    end
+    get image_url(-1)
+    assert_response :not_found
```
bboe: The way this was done before is acceptable.

test/controllers/images_controller_test.rb
```diff
@@ -82,4 +81,25 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
   end
+
+  test 'should delete image and should not show that image' do
+    image = Image.create!(link: @sample_img_link)
+    assert_difference('Image.count', -1) do
+      delete image_url(image)
+    end
+
+    assert_redirected_to images_url
+    assert_select "img[src='#{image.link}']", false
```
bboe: I don't think this does what you think it does. There should be no body on a redirect.

test/controllers/images_controller_test.rb
```diff
@@ -82,4 +81,25 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    assert_equal 'Image was successfully destroyed.', flash[:success]
+
+    get image_url(image)
```
bboe: Each controller test should only have only one get/post/put/delete method call. This check is redundant with the `assert_difference`.

test/controllers/images_controller_test.rb
```diff
@@ -82,4 +81,25 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    assert_response :not_found
+  end
+
+  test 'should not delete image with invalid id' do
+    assert_difference('Image.count', 0) do
```
bboe: FYI, you can also use `assert_no_difference`.

test/flows/image_flow_test.rb
```diff
@@ -4,19 +4,29 @@ class ImageFlowTest < FlowTestCase
   test 'add new image with no tags' do
```
bboe: Does this necessitate its own flow? Adding without tags probably can be handled just by the controller.

test/flows/image_flow_test.rb
```diff
@@ -4,19 +4,29 @@ class ImageFlowTest < FlowTestCase
...
     fill_in('image[link]', with: 'http://i.imgur.com/PVrX7wZ.jpg')
-    find('input[type="submit"]').click
+    click_button 'Create Image'
     assert has_current_path?(image_path(Image.last))
-    assert_equal 'Image was successfully created.', find('#notice').text
+    assert_equal 'Image was successfully created.', find('#success').text
     assert_equal Image.last.link, find('img')[:src]
   end
 
+  test 'add new image then delete' do
```
bboe: This is a great example of a flow that a user may do.

test/flows/image_flow_test.rb
```diff
@@ -4,19 +4,29 @@ class ImageFlowTest < FlowTestCase
...
+    visit(new_image_path)
+    fill_in('image[link]', with: 'http://i.imgur.com/PVrX7wZ.jpg')
+    find('input[type="submit"]').click
+    assert has_current_path?(image_path(Image.last))
+    click_link 'Destroy'
+    assert_equal 'Image was successfully destroyed.', find('#success').text
+    assert has_current_path?(images_path)
+  end
+
   test 'add new image with single tags' do
```
bboe: Also something that doesn't require a flow test. While many of the flow-test components overlap with controller tests, we should avoid tests that only overlap with controller tests.

bboe:
@mqchau I added some comments. Please create a new pull request to address them. I'll create an associated issue so you have a placeholder for this comment.

Also please get in the habit of deleting your branch after merging:

image

app/controllers/images_controller.rb
mqchau: I just find out that Rails behaves differently on production vs development. 

In development mode, using Image.find would result in ActiveRecord::RecordNotFound exception. In production mode, using Image.find will return HTTP status code 404 and 404 page.

Should I use Image.find and then make all my test assert_raises  ActiveRecord::RecordNotFound?

app/views/layouts/application.html.erb
mqchau: Stanley and I moved it to images scope. I'm going to move it back.

test/controllers/images_controller_test.rb
mqchau: It seems that minitest automatically GET the redirected page after assert_redirected_to. If I do a separate "get images_url" after that assert it would work too. Do you think it's necessary to be that explicit?

test/flows/image_flow_test.rb
mqchau: I and Stanley thought that it wouldn't hurt to add its own flow.

app/controllers/images_controller.rb
bboe: > Should I use Image.find and then make all my test assert_raises ActiveRecord::RecordNotFound?

Yes, this is the same in production and in dev. The difference is the middlewear (at least I think that's where it is) that converts RecordNotFound to a 404. This only occurs in production.

test/controllers/images_controller_test.rb
bboe: I think because you have the false there at the end it actually passes this assertion because it doesn't exist in the body. Hence why I think it doesn't do what you think it does.

test/flows/image_flow_test.rb
bboe: I'm going to disagree. Flow tests should not cover the same thing as controller tests. If you can test it in a controller test, then that's what you should do. The flow tests should be for ensuring that the transition from page to page, data entry, and user-interface works as expected. This test does not seem to add anything other than test time.

test/controllers/images_controller_test.rb
mqchau: "puts response.body" after assert_redirected_to shows the body of the redirect message. Apparently I need to do a get to the link to get the correct body.

bboe: What I'm saying is that this is unnecessary. This test should be `should delete image` and only assert that you are redirected when done. There's no reason to check the redirected page response body.

mqchau: I think it is necessary that we check to see if the image appears again after the delete action. The issue said "Deleted images are removed from the homepage.".

bboe: Yes, the issue does say that. That test would be appropriate for the flow test. Or during QA. But it doesn't belong as a separate request in the controller test. Also it's worth noting that `assert_difference` already ensures that an image is deleted. You can use flow tests to ensure the right one was deleted.

PR #29, opened 2016-07-26
add 20 images in db seed file

### LINKED ISSUE #5: Image Seeds
#### As a developer I want at least 20 images to be created when I setup the database.

Acceptance criteria:
- [x] After running `rake db:setup` at least 20 images are shown in the
  homepage.
- [x] When a new app is deployed on heroku (e.g., every review app), that app
  automatically has at least 20 images on its homepage.

Dependencies:
- Image Index

mqchau:
Closes #5


mlewisno:
Looks good overall but two questions:
1) What did you have to do in order to get the images to appear after clicking he deploy button? is there a way you could automate that process?
2) I also noticed in the diff that you chose to use Image.create instead of the old Image.create! Can you tell me a bit about what the difference is and why you made the choice you did?

mlewisno:
And please don't merge until we've discussed those questions more ;)

mqchau:
1) I don't have to do anything. The deployment on review apps on Heroku automatically does db:seed because I have that in app.json. Only in staging app and production app I have to manually run db:migrate and db:seed. But that is expectable.

2) That's good catch. I missed that due to merging between branches. Fixing it now.

mlewisno:
For 1 - check in with Bryce but it was my understanding (as of mid day yesterday) that one of the goals of this particular exercise was to make the db:migrate and db:seed run automatically on stage and production as well - but if I'm mistaken please let me know so I don't make the same mistake 😄

mqchau:
Ok asked him via Slack. Will post response here.

mlewisno:
Excellent! Thanks!

mqchau:
from Bryce: 
Once an app is deployed you have to run migrations manually.
Only newly launched applications do anything with `app.json`.

So I think we can't / aren't supposed to run db:seed automatically on staging and production app on Heroku. 

As a result, I think this commit suffice the requirements in the issue.

PR #32, opened 2016-07-28
Qc tag image search

### LINKED ISSUE #9: Images by Tag
#### As a user I want to view all images associated with a tag.

Acceptance criteria:
- [ ] When I click on a tag, I see a filtered list of only the images that have
  the tag I clicked on.
- [ ] The tag name (not its ID) is used to perform the filtering.

Discussion topics:
- [ ] Where should we implement this functionality?
- [ ] What happens if a user tries to manually filter by a nonexistent tag?

Dependencies:
- Image Tags

mqchau:
Closes #9


app/controllers/images_controller.rb
```diff
@@ -28,6 +28,11 @@ def destroy
     redirect_to images_url
   end
 
+  def tagshow
+    @current_tag = params[:name]
+    @images_by_tag = Image.tagged_with([@current_tag], any: true)
+  end
+
```
shiboying: To me, it is ok to show images in a new page. But could you think another way to show images just on image#index page? So, you could think the tag is just like a filter. When you view index page, the image#index page basically show all the images. And imaging you hit a tag, what we want is basically show fewer images with this tag. Sill like a index page right? So could you change to that way?

shiboying: Hint: `http://localhost:3000/images?tag=abc`

app/views/images/index.html.erb
```diff
@@ -1,8 +1,11 @@
 <h1>Images</h1>
-
 <% @images.each do |image| %>
   <%= image_tag(image.link, :class => 'image__medium_display') %>
-  Tags: <p class="image__tag"><%= image.tag_list %></p>
+  Tags:
+    <% image.tag_list.each do |image_tag| %>
+        <%= link_to image_tag, image_tag_show_url(image_tag), class: 'image__tag' %>
```
shiboying: Could you explain why you use `image_tag_show_url(image_tag)` instead of `image_tag_show_path(image_tag)`? And why give it a class `image__tag`?

app/views/images/show.html.erb
```diff
@@ -1,6 +1,10 @@
 <p>
   <%= image_tag(@image.link) %>
-  Tags: <p class="image__tag"><%= @image.tag_list %></p>
+  Tags:
+  <% @image.tag_list.each do |image_tag| %>
+      <%= link_to image_tag, image_tag_show_url(image_tag), class: 'image__tag' %>
+  <% end %>
+
```
shiboying: In APM, we usually DRY our code. Since this bunch of code is the same with what it is in index page, you should extract it to a partial to render this page. Also, after you render it to a partial, you could think a way to avoid using the `for each` loop.

app/views/images/tagshow.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if @images_by_tag.any? %>
+  <h1>Images tagged <%= @current_tag %></h1>
+  <% @images_by_tag.each do |image| %>
+      <%= image_tag(image.link, :class => 'image__medium_display') %>
+      <br/>
+  <% end %>
+<% else %>
+    <h1>No images is tagged <%= @current_tag %></h1>
+<% end %>
+
```
shiboying: If you can filter images on image#index page, you could avoid creating a view page for it.

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
   resources :images, only: [:show, :create, :new, :index, :destroy]
+  get '/images/tag/:name', to: 'images#tagshow', as: 'image_tag_show'
```
shiboying: The same, you could also avoid using this.

test/controllers/images_controller_test.rb
```diff
@@ -4,6 +4,7 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
   setup do
     @sample_img_link = 'http://i.imgur.com/PVrX7wZ.jpg'
     @sample_img_tags = 'testtag1, testtag2'
+    @sample_img_tags_array = @sample_img_tags.split(',').collect(&:strip)
```
shiboying: If you don't use `@sample_img_tags`, you could just give a setup for the array, instead of doing the split on the string. You could just use this: `@sample_img_tags = ['testtag1', 'testtag2']`

test/controllers/images_controller_test.rb
```diff
@@ -59,7 +60,9 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     get image_url(image)
     assert_response :success
     assert_select 'img[src=?]', image.link
-    assert_select '.image__tag', @sample_img_tags
+    @sample_img_tags_array.each do |x|
```
shiboying: If you could give a more descriptive name rather than `x`, that is more desirable.

test/controllers/images_controller_test.rb
```diff
@@ -74,7 +77,9 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
     assert_select 'h1', 'Images'
     assert_select 'img[src=?]', image.link
-    assert_select '.image__tag', @sample_img_tags
+    @sample_img_tags_array.each do |x|
```
shiboying: The same.

test/controllers/images_controller_test.rb
```diff
@@ -74,7 +77,9 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+      assert_select '.image__tag', x
```
shiboying: I see the point you give a class name, you can just ignore my comment on that.

test/controllers/images_controller_test.rb
```diff
@@ -100,4 +105,15 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+
+  test 'should show images by tag' do
+    image1 = Image.create!(link: @sample_img_link, tag_list: @sample_img_tags)
+    image2 = Image.create!(link: 'http://i.imgur.com/mWwsvks.jpg', tag_list: @sample_img_tags_array[0])
+    get image_tag_show_url(@sample_img_tags_array[0])
+    assert_select 'img[src=?]', image1.link
+    assert_select 'img[src=?]', image2.link
+    get image_tag_show_url(@sample_img_tags_array[1])
+    assert_select 'img[src=?]', image1.link
+    assert_select "img[src='#{image2.link}']", false
```
shiboying: Better to give a new line between setting test and assert test result.

test/flows/image_flow_test.rb
```diff
@@ -16,4 +16,27 @@ class ImageFlowTest < FlowTestCase
...
+    visit(new_image_path)
+    fill_in('image[link]', with: sample_img_link2)
+    fill_in('image[tag_list]', with: tag_name)
+    find('input[type="submit"]').click
+    assert has_current_path?(image_path(Image.last))
+    assert_equal 'Image was successfully created.', find('#success').text
+    assert_equal Image.last.link, find('img')[:src]
+    visit(image_tag_show_path(tag_name))
+    page.assert_selector("img[src='#{sample_img_link1}']")
+    page.assert_selector("img[src='#{sample_img_link2}']")
```
shiboying: The same, if without new line, it is hard to read the code for future developers.

mqchau:
I think my way is more RESTful and Sebastian agrees with me. I think we should leave it this way.

app/views/images/show.html.erb
mqchau: So i moved this part to a partial template. But I don't know how to avoid a for loop. What would you suggest?

shiboying:
@mqchau https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/9#discussion_r51065358 You can read this discussion together with Seb.

app/views/images/show.html.erb
shiboying: You can read this: http://guides.rubyonrails.org/layouts_and_rendering.html `3.4.5 Rendering Collections`

app/controllers/images_controller.rb
shiboying: Also, I think it is necessary to consider the case if there is no such tag. What if someone just manually change the url to be `/images/tag/malicioustag`( in your way) `/images/index?tag=malicioustag`. What page you will render? And also you need to write a test for this case.

mqchau:
I think I have addressed all your concerns. Please take a look when you can.

app/controllers/images_controller.rb
```diff
@@ -2,7 +2,9 @@ class ImagesController < ApplicationController
   before_action :set_image, only: [:show, :destroy]
 
   def index
-    @images = Image.all
+    @tag = params[:tag]
+    @tag = @tag.split(',')[0] unless @tag.nil?
+    @images = @tag.nil? ? Image.all : Image.tagged_with([@tag], any: true)
```
shiboying: You check `@tag.nil?` twice. Could you think a way to just use one to do the logic here? And if there is no such tag, I think it would be better to add a flash message here to warn user what is going on here.

app/views/images/index.html.erb
```diff
@@ -1,8 +1,7 @@
-<h1>Images</h1>
-
+<h1><%= "No" if @images.empty? %> Images <%= "tagged #{@tag}" if @tag.present? %></h1>
 <% @images.each do |image| %>
   <%= image_tag(image.link, :class => 'image__medium_display') %>
-  Tags: <p class="image__tag"><%= image.tag_list %></p>
+  <%= render partial: 'tagshow', collection: image.tag_list, as: :tag %>
```
shiboying: If you do like this: ` <%= render partial: 'tag', collection: image.tag_list %>` you can directly use `tag` as the local variable in the partial to refer each tag in the tag_list (if you name it `_tag.html.erb` I think in this way, it would be nice and clean.

app/views/images/show.html.erb
```diff
@@ -1,6 +1,6 @@
 <p>
   <%= image_tag(@image.link) %>
-  Tags: <p class="image__tag"><%= @image.tag_list %></p>
+  <%= render partial: 'tagshow', collection: @image.tag_list, as: :tag %>
+
```
shiboying: Same here.

mqchau:
Yeah I agree the logic in checking tag params was sloppy. I fixed it to be cleaner. I also change partial template name to _tag as you said.

shiboying:
Good job! It looks good to me.

mqchau:
Ok I'll merge it then.

PR #33, opened 2016-07-28
require images to have at least one tag to create

### LINKED ISSUE #12: Required Tag Association
#### As a user I want all images to have at least one tag.

Acceptance criteria:
- [x] I cannot save an image without adding at least one tag.
- [x] An error message is associated with the appropriate input field when a
  tag is not provided.

Dependencies:
- Image Tags

mqchau:
Closes #12


app/models/image.rb
```diff
@@ -1,4 +1,12 @@
 class Image < ApplicationRecord
   validates :link, url: true
+  validate :tag_exist
```
pololee: Since you are using `ActiveRecord`, you can use `validates_presence_of`.
http://api.rubyonrails.org/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_presence_of

app/models/image.rb
```diff
@@ -1,4 +1,12 @@
...
+
   acts_as_taggable
+
+  private
+
+  def tag_exist
```
pololee: With `validates_presence_of`, you don't need this method.

test/models/image_test.rb
```diff
@@ -6,32 +6,42 @@ class ImageTest < ActiveSupport::TestCase
     @sample_img_link = 'http://i.imgur.com/PVrX7wZ.jpg'
   end
 
-  test 'create valid image with no tag' do
-    image = Image.new(link: @sample_img_link)
-    assert_predicate image, :valid?
+  test 'should not create valid image with no tag' do
+    image = Image.new(link: @sample_img_link, tag_list: '')
+    refute_predicate image, :valid?
```
pololee: You might also want to test the error message.

test/models/image_test.rb
```diff
@@ -6,32 +6,42 @@ class ImageTest < ActiveSupport::TestCase
...
+  end
+
+  test 'should not create valid image with nil tag' do
+    image = Image.new(link: @sample_img_link, tag_list: nil)
+    refute_predicate image, :valid?
```
pololee: You might also want to test the error message.

test/models/image_test.rb
```diff
@@ -6,32 +6,42 @@ class ImageTest < ActiveSupport::TestCase
...
   end
 
   test 'invalid image URL with no tag' do
-    image = Image.new(link: 'httpwww.abc.com/cde')
+    image = Image.new(link: 'httpwww.abc.com/cde', tag_list: 'testtag1')
```
pololee: The test name is `invalid image URL with no tag`. But you actually assigned a tag. You may want to fix it.

test/models/image_test.rb
```diff
@@ -6,32 +6,42 @@ class ImageTest < ActiveSupport::TestCase
...
-    image = Image.new(link: nil)
+    image = Image.new(link: nil, tag_list: 'testtag1')
     refute_predicate image, :valid?
     assert_predicate image.errors, :any?
     assert_includes image.errors.full_messages, @invalid_img_url_err_msg
   end
 
+  test 'should not create valid image with no tag and invalid URL' do
+    image = Image.new(link: 'httpwww.abc.com/cde', tag_list: '')
+    refute_predicate image, :valid?
```
pololee: You might also want to test the error messages.

pololee:
I checked your [`ImageFlowTest`](https://github.com/AppFolioOnboarding/image-sharer-mqchau/blob/master/test/flows/image_flow_test.rb). You only have a test case for everything goes right. You might want to add a test for sth. goes wrong, like no tag, invalid url.

mqchau:
On another issue I was doing similar thing as you said. And Bryce said that those cases are covered in controller test already. So we don't have to cover them in flow test. I also agree on that.

bboe:
I don't think there should be a separate test for something goes wrong. However, I do think there should at least one "oops I typed something in correctly" as part of the flow.

mqchau:
I just learn some fundamental things of Rails while adding to the flow the action that user forgot to add a tag.

The controller for create I wrote is:

```ruby
  # GET /images/new
  def create
    @image = Image.new(image_params)
    if @image.save
      flash[:success] = 'Image was successfully created.'
      redirect_to @image
    else
      render :new, status: :unprocessable_entity
    end
  end
```

When the save fails, the URL on the browser will be /images instead of /images/new. Is this the behavior we want? I find it confusing that first time the URL is /images then 2nd time it's /images/new when user enters wrong info.

bboe:
> When the save fails, the URL on the browser will be /images instead of /images/new.

That is expected. We're POSTing to /images for the creation, and it's convenient to return the form with errors as a reply as part of that request rather than redirecting to `/images/new` with the invalid form data.

mqchau:
My latest commit added to the flow test a small section that the user enter a valid URL but no tag. Then he enters some tag and it works. Then he deletes the image.

pololee:
Sorry, I forgot to watch this repo. I agree with Bryce and I should have clarified my comment.
I took a look at the latest commit. It's good. Thx.

test/models/image_test.rb
```diff
@@ -6,27 +6,34 @@ class ImageTest < ActiveSupport::TestCase
...
+  test 'should not create valid image with nil tag' do
+    image = Image.new(link: @sample_img_link, tag_list: nil)
+    refute_predicate image, :valid?
+    assert_equal "can't be blank", image.errors.messages[:tag_list].first
   end
 
-  test 'invalid image URL with no tag' do
-    image = Image.new(link: 'httpwww.abc.com/cde')
+  test 'invalid image URL with valid tag' do
+    image = Image.new(link: 'httpwww.abc.com/cde', tag_list: 'testtag1')
```
bboe: Notice that because you made tag_list required you have to update all your tests to add the tag list parameter. How can you refactor this such that you need only update one line whenever you add a required parameter?

mqchau: I would do this:

```ruby
setup do
  @sample_params = { link: '____', tag_list: 'tag1, tag2' }
end

test 'invalid tag' do
  params = @sample_params.clone
  params[:tag_list] = '' 
  Image.create!(params)
  # assert something
end
```

Is this what you mean?

bboe: I was thinking something more along the lines of a private factory function `create_image` with default arguments such that you need only call `create_image` in order to always get a valid image, and you can overwrite the values by adjusting the arguments to `create_image`.

pololee: I would highly recommend `keyword Argument` for this private function.
https://robots.thoughtbot.com/ruby-2-keyword-arguments

mqchau:
I think I got what you guys mean.

mqchau:
Some how the passed result of CircleCI doesn't get to this PR on github. This is strange.

mqchau:
So can I merge this PR now that I addressed your last comment, @bboe ?

bboe:
👍

PR #34, opened 2016-08-02
add image edit page to modify tags of an image

### LINKED ISSUE #13: Modify Tag List
#### As a user I want to edit the tag list for an image.

Acceptance criteria:
- [ ] There is a link to edit the tags for an image.
- [ ] The image whose tags are to be modified is displayed on the edit page.
- [ ] On the edit page, I can change the tags to any valid tag list.
- [ ] After saving the edit page with valid tags, my changes are visible
  throughout the rest of the application.

Dependencies:
- Image Tags

mqchau:
Closes #13


app/controllers/images_controller.rb
```diff
@@ -36,6 +36,18 @@ def destroy
...
 
+  def edit
+  end
+
+  def update
+    if @image.update(image_tag_params)
+      flash[:success] = 'Image tag list was successfully modified.'
+      redirect_to @image
+    else
+      render :edit, status: :unprocessable_entity
```
weixing2014: 👍

app/views/images/_form.html.erb
```diff
@@ -12,8 +12,10 @@
   <% end %>
 
   <div class="field">
-    <%= f.label :link %>
-    <%= f.text_field :link %>
+    <% if curr_action == 'new' %>
+      <%= f.label :link %>
+      <%= f.text_field :link %>
+    <% end %>
```
weixing2014: Is it possible that a user wants to update the link?

mqchau:
Since you have DM me on Slack that this PR is OK, I'm going to merge it now. Thank you for reviewing it.

app/controllers/images_controller.rb
```diff
@@ -36,6 +36,18 @@ def destroy
     redirect_to images_url
   end
 
+  def edit
+  end
+
+  def update
+    if @image.update(image_tag_params)
```
bboe: Do you want the user to be able to edit the url?

mqchau: Does the function image_tag_params suffice? It only accepts tag_list argument and nothing else.

bboe: It also accepts `link` which means that the link can be modified.

app/controllers/images_controller.rb
```diff
@@ -45,4 +57,8 @@ def set_image
   def image_params
     params.require(:image).permit(:link, :tag_list)
   end
+
+  def image_tag_params
+    params.require(:image).permit(:tag_list)
+  end
```
mqchau: It doesn't accept link. I have two separate functions to parse post params: one for new and one for update

bboe: Sorry, I missed that. I was looking at image_params. Yes. that is sufficient. At this point I am curious if the `image_params` functions are necessary if they are only used in one location.

mqchau: I think it's OK to leave it this way.

bboe: 👍

PR #35, opened 2016-08-02
add tag index page

### LINKED ISSUE #16: Tag Index
#### As a user I want to see a list of tags and the number of images associated with each.

Acceptance criteria:
- [ ] There is a link in the application that takes me to a list of all tags.
- [ ] The tag list is sorted by frequency with the most common first.
- [ ] Each tag name is a link to the image list filtered by that tag.

Dependencies:
- Images by Tag

mqchau:
Closes #16


app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,7 @@
+class TagsController < ApplicationController
+  def index
+    @tag_count = ActsAsTaggableOn::Tag.all
```
bboe: Can you use `ActsAsTaggableOn::Tag.most_used` for this?

bboe: Does `tag_count` accurately represent what this variable holds? I would expect that to be an integer.

app/views/tags/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<% @tag_count.each do |t| %>
```
bboe: Can we use a more descriptive name than `t`?

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
   resources :images
   root 'images#index'
+  get '/tags', to: 'tags#index'
```
bboe: Can you use the `resources` method for listing routes? It makes it easier to update later to add more actions to the controller if we were to do that.

test/controllers/tags_controller_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'test_helper'
+
+class TagsControllerTest < ActionDispatch::IntegrationTest
+  test 'should get tag index' do
```
bboe: What happens if there are no tags?

bboe:
I added some feedback in @ipmsteven's absence.

app/views/layouts/application.html.erb
```diff
@@ -10,6 +10,7 @@
 
   <body>
 
+    <%= link_to 'All tags', tags_path %>
```
ipmsteven: do we want have `all tags` link on every pages?

mqchau: Yeah my plan is to have that link on the top navigation bar. And the navbar is already done in separate branch for styling. Will merge these two branches together.

app/views/tags/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<% @tags.each do |tag| %>
+  <p>
+    <%= render 'images/tag', tag: tag[:name] %>
+    <span class="tag__count"><%= tag[:taggings_count] %></span>
+  </p>
```
ipmsteven: _ProTip_:  [Rendering Collections in Rails](https://robots.thoughtbot.com/rendering-collections-in-rails)

app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,5 @@
+class TagsController < ApplicationController
+  def index
+    @tags = ActsAsTaggableOn::Tag.most_used
```
ipmsteven: _Food for thoughts_: http://blog.arkency.com/2016/07/respond-to-format-is-useful-even-without-multiple-formats/

app/views/layouts/application.html.erb
ipmsteven: ok

test/controllers/tags_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+class TagsControllerTest < ActionDispatch::IntegrationTest
+  test 'should get tag index no tags' do
+    get tags_path
+
+    assert_response :success
+    assert_select '.image__tag', false
+    assert_select '.tag__count', false
+  end
+
+  test 'should get tag index with some tags' do
```
ipmsteven: could we also check the each tag has the correct count?

mqchau: Yeah that's the next test.

app/controllers/tags_controller.rb
bboe: Thanks for sharing! I wasn't aware that those 500 errors would occur. I guess we should suggest using `respond_to` from the beginning. Although I wonder if there is a way to 406 any non-html requests so respond_to isn't necessary.

ipmsteven: yeah, someone mentioned in the comments from the article to use a middleware to 406 any non-html requests

mqchau: If I use respond_to that means I would have to redo all of existing controller. I think this task can be a separate issue and I'll deal with it later.

mqchau:
Hi Yunlei, do you think this PR is good to merge yet?

mqchau:
After conversation on Slack with Yunlei, I will merge this PR.

PR #36, opened 2016-08-03
add style

### LINKED ISSUE #6: Add Style
#### As a user I want to see a visually pleasing application.

Acceptance criteria:
- [x] The application has a header with navigation
- [x] Images are displayed in an aesthetically pleasing list of cards
- [x] Form errors are shown inline with the related input field
- [x] I know how to use Bootstrap components like buttons and navbars in my
  application.
- [x] I know the very basics of CSS and SCSS

Reference material:
- http://v4-alpha.getbootstrap.com/getting-started/introduction/
- https://github.com/plataformatec/simple_form

Dependencies:
- Image Index

mqchau:
Closes #6


app/assets/stylesheets/application.scss
```diff
@@ -13,3 +13,8 @@
  *= require_tree .
  *= require_self
  */
+
```
aaronpanch: If we are using SCSS imports, we don't want the Sprockets requires above.  It's an either/or thing.

app/assets/stylesheets/images.scss
```diff
@@ -0,0 +1,13 @@
+.image__medium_display {
+  max-width: 400px;
+  width: 100%;
+  padding-bottom: 10px;
+}
+
+.image__tag {
```
aaronpanch: The naming convention here indicates that this is a subcomponent of the "image" component; however, I don't see a CSS rule defined for one, nor a usage in the markup.

app/assets/stylesheets/images.scss
```diff
@@ -0,0 +1,13 @@
...
+  display: inline;
+}
+
+.image__action_button_group {
```
aaronpanch: Perhaps this should be it's own component?

app/assets/stylesheets/images.scss
```diff
@@ -0,0 +1,13 @@
+.image__medium_display {
```
aaronpanch: This class also seems to be dangling, and doesn't belong to a parent component.  These seem OK, they just need a little naming change.

app/assets/stylesheets/images.scss
```diff
@@ -0,0 +1,13 @@
...
+  max-width: 400px;
+  width: 100%;
+  padding-bottom: 10px;
```
aaronpanch: Also, it's nice to alphabetize CSS properties to make it easy to read.

app/assets/stylesheets/scaffolds.scss
```diff
@@ -0,0 +1,73 @@
+body {
```
aaronpanch: Do you want to use the scaffolds file? Using bootstrap circumvents this stylesheet, and I suggest using one "library".

app/views/images/index.html.erb
```diff
@@ -1,8 +1,13 @@
...
-  <%= image_tag(image.link, :class => 'image__medium_display') %>
-  <%= render partial: 'tag', collection: image.tag_list %>
-  <%= link_to 'Destroy', image, method: :delete, data: { confirm: 'Are you sure?' } %>
-  <%= link_to 'Edit', edit_image_path(image) %>
-  <br/>
-<% end %>
+<h1 class="h1">Images <%= "tagged #{@tag}" if @tag.present? %></h1>
+<div class="card-group">
+  <% @images.each do |image| %>
+    <div class="well well-sm card col-md-4 image__box">
```
aaronpanch: I don't see a CSS rule defined for image__box.

aaronpanch:
It would be nice to add some space between each of the image "cards" here:
screen shot 2016-08-03 at 10 15 46 am

aaronpanch:
Upon adding an image, I noticed it put the latest image at the bottom, could you reverse the order so the most recently added images appear at the top?

app/views/images/_form.html.erb
```diff
@@ -1,26 +1,28 @@
-<%= form_for(image) do |f| %>
-  <% if image.errors.any? %>
-    <div id="error_explanation">
-      <h2><%= pluralize(image.errors.count, "error") %> prohibited this image from being saved:</h2>
+<div class="col-md-4">
+  <%= form_for(image) do |f| %>
```
aaronpanch: I might have forgotten to mention this, but checkout the gem [simple_form](https://github.com/plataformatec/simple_form).  It makes creating forms super easy, and integrates with bootstrap to make errors look great too.

app/assets/stylesheets/images.scss
mqchau: It's used in _tag.html.erb . And isn't this the CSS rule?

app/assets/stylesheets/scaffolds.scss
mqchau: Ok I'll try to extract only a few things I need from scaffold. It got some colorful error display.

app/assets/stylesheets/images.scss
mqchau: I think this is ok, these button group applies to image only.

app/assets/stylesheets/images.scss
mqchau: I don't understand what you suggest here.

app/views/images/_form.html.erb
mqchau: I think the way I'm doing is OK, it looks like this

image

mqchau:
Hi Aaron, does this commit look good for you to merge?

app/views/images/_form.html.erb
aaronpanch: We use simple_form in APM and we should use it for this project to gain familiarity etc. In addition, it helps reduced the number of styles we have to write.  Namely, the file `form_common.scss` is not needed when using bootstrap, and simple forms helps us make form elements interface with bootstrap.

app/assets/stylesheets/images.scss
aaronpanch: Sorry, you're right, the comment isn't clear! What I mean is that the names in this file don't follow our naming convention.  Every file corresponds to a "component" which could be one element on the page or a configuration of elements on the page.  For the styles listed here, what is the "component" in question? Once we figure that out, then we can name it's pieces accordingly.

mqchau:
Ok I will use simple form.

app/assets/stylesheets/application.scss
aaronpanch: I see you added a +1 to the comment, but the sprockets statements are still there.

app/controllers/images_controller.rb
```diff
@@ -4,10 +4,10 @@ class ImagesController < ApplicationController
   def index
     raw_tag_param = params[:tag]
     if raw_tag_param.nil?
-      @images = Image.all
+      @images = Image.all.reverse_order
```
aaronpanch: Perhaps we should be explicit about the order, such as order by created at desc

app/assets/stylesheets/footer.scss
```diff
@@ -0,0 +1,20 @@
...
+.footer {
+  position: absolute;
+  bottom: 0;
+  width: 100%;
+  /* Set the fixed height of the footer here */
+  height: 60px;
+  background-color: #f5f5f5;
+}
+
+.footer p {
```
aaronpanch: styling using element selectors can be brittle because it's tied to the markup.  A more robust solution is to come up with a class to assign to each of these `p` elements.  Don't forget to follow the naming convention.

app/views/images/index.html.erb
```diff
@@ -1,8 +1,19 @@
...
-<% @images.each do |image| %>
-  <%= image_tag(image.link, :class => 'image__medium_display') %>
-  <%= render partial: 'shared/taglink', collection: image.tag_list %>
-  <%= link_to 'Destroy', image, method: :delete, data: { confirm: 'Are you sure?' } %>
-  <%= link_to 'Edit', edit_image_path(image) %>
-  <br/>
-<% end %>
+<h1 class="h1">Images <%= "tagged #{@tag}" if @tag.present? %></h1>
+<div class="card-group">
+  <% Array(0..2).each do |column_index| %>
```
aaronpanch: I suspect this logic to group cards into three is caused by the weird alignment issues when throwing all the cards into one "row".   If we enable flexbox grids in bootstrap, that actually fixes this problem!  It's as simple as adding the `$enable-flex: true;` variable to the top of your manifest.  Then, this logic can be cleaned up.

app/views/images/show.html.erb
```diff
@@ -1,7 +1,12 @@
 <p>
-  <%= image_tag(@image.link) %>
-  <%= render partial: 'shared/taglink', collection: @image.tag_list %>
-
-  <%= link_to 'Edit Tag List', edit_image_path(@image) %>
-  <%= link_to 'Destroy', @image, method: :delete, data: { confirm: 'Are you sure?' } %>
+  <div class="row">
```
aaronpanch: I don't think `div` elements can go inside a `p`. It must be phrasing content: https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2, or you can remove the `p`--i'm not sure.

app/views/images/index.html.erb
```diff
@@ -1,8 +1,19 @@
...
+    <div class="col-md-4">
+      <% @images.each_with_index do |image, index| %>
+        <% if index % 3 == column_index %>
+          <div class="card image__box">
```
aaronpanch: Where are the styles for this?

config/initializers/simple_form.rb
```diff
@@ -0,0 +1,164 @@
+# Use this setup block to configure all options available in SimpleForm.
+SimpleForm.setup do |config|
```
aaronpanch: You only need one of these initializers (not this one, the bootstrap one)

app/views/layouts/application.html.erb
```diff
@@ -10,8 +10,11 @@
 
   <body>
 
-    <%= link_to 'All tags', tags_path %>
-    <%= render partial: 'layouts/flash_message' %>
-    <%= yield %>
+    <%= render partial: 'layouts/navbar' %>
+    <div class="container wrapper">
```
aaronpanch: where's the wrapper class defined?

app/assets/stylesheets/application.scss
```diff
@@ -0,0 +1,8 @@
+$brand-success : #5aca52;
+@import "bootstrap";
+@import "images/*";
+@import "footer";
+
+body {
```
aaronpanch: I see you've noticed that when we add a fixed nav-bar it covers up part of the page.  This is great.  One thing: I don't like to put styles directly in the application manifest -- therefore, we can create a file called `_base.scss` and move any base styles there.  Be sure to import it before your components but after the bootstrap library.

config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,151 @@
...
+
+  config.wrappers :vertical_file_input, tag: 'div', class: 'form-group', error_class: 'has-error' do |b|
+    b.use :html5
+    b.use :placeholder
+    b.optional :maxlength
+    b.optional :readonly
+    b.use :label, class: 'control-label'
+
+    b.use :input
+    b.use :error, wrap_with: { tag: 'span', class: 'help-block' }
```
aaronpanch: Changing the class here will fix the error styling for forms.

config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,151 @@
+# Use this setup block to configure all options available in SimpleForm.
+SimpleForm.setup do |config|
+  config.error_notification_class = 'alert alert-danger'
+  config.button_class = 'btn btn-default'
+  config.boolean_label_class = nil
+
+  config.wrappers :vertical_form, tag: 'div', class: 'form-group', error_class: 'has-error' do |b|
```
aaronpanch: Also the `has-error` class can be changed here too.

app/assets/stylesheets/_footer.scss
```diff
@@ -0,0 +1,7 @@
+.footer {
+
+}
+
+.footer__text {
```
aaronpanch: I would rather put padding on the `.footer` parent component, so that it has a style, and then we probably won't need this subcomponent.

app/assets/stylesheets/_image-box.scss
```diff
@@ -0,0 +1,12 @@
+.image-box {
```
aaronpanch: Usually components that don't have any style have a _bad smell_; however, here is seems that you're using the existing bootstrap class so I'll let it slide.  Can you think of any way this could be rearranged so all classes have some style?

app/views/images/_imagebox.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="card image-box">
+  <a href="<%= image_url(image) %>"><%= image_tag(image.link, class: 'image-box__image card-img-top') %></a>
+  <div class="card-block">
+    <%= render partial: 'shared/taglink', collection: image.tag_list %>
+  </div>
+  <div class="card-block image-box__actions">
```
aaronpanch: Is `image-box__actions` still used?

app/assets/stylesheets/_image-box.scss
aaronpanch: I wonder if the component is really necessary now.  Perhaps the component really should just be the .image-box__image, and the .image-box__tags should be a utility class: http://v4-alpha.getbootstrap.com/components/utilities/#css-display-block-inline-inline-block

mqchau: I need this so I can select the .image-box from flow test / controller test

aaronpanch: Ah, if that's the case.  We typically use th `.js-` prefix for Javascript or tests.  That way, the markup/styles are decoupled from the tests.  So, you can replace `image-box` with `js-image-box` and that way the test is only tied to that class. If we wanted to add style in the future, it wouldn't break the flow test.

app/views/images/_form.html.erb
```diff
@@ -1,26 +1,9 @@
...
-      <ul>
-      <% image.errors.full_messages.each do |message| %>
-        <li><%= message %></li>
-      <% end %>
-      </ul>
-    </div>
-  <% end %>
-
-  <div class="field">
+<div class="col-md-4">
```
aaronpanch: Is this contained in a element with `.row` somewhere -- I can't find it and I think bootstrap needs columns to be contained by a row.

app/views/images/_imagebox.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="card image-box">
+  <a href="<%= image_url(image) %>"><%= image_tag(image.link, class: 'image-box__image card-img-top') %></a>
```
aaronpanch: Not a big deal, but in ruby you don't need the parenthesis for function invocation (unless there's composition going on)

app/views/images/edit.html.erb
```diff
@@ -1,4 +1,6 @@
-<h1>Edit image</h1>
+<h1>Edit image tags</h1>
 
-<%= image_tag(@image.link) %>
+<div class="col-md-4">
```
aaronpanch: Same comment here about there needing to be a parent `.row`

app/views/images/show.html.erb
```diff
@@ -1,7 +1,3 @@
-<p>
-  <%= image_tag(@image.link) %>
-  <%= render partial: 'shared/taglink', collection: @image.tag_list %>
-
-  <%= link_to 'Edit Tag List', edit_image_path(@image) %>
-  <%= link_to 'Destroy', @image, method: :delete, data: { confirm: 'Are you sure?' } %>
-</p>
+<div class="col-md-4">
```
aaronpanch: `.row` here.

app/views/layouts/_navbar.html.erb
```diff
@@ -0,0 +1,16 @@
+<nav class="navbar navbar-dark navbar-fixed-top bg-inverse">
+  <div class="container">
+    <a class="navbar-brand" href="<%= root_path %>"><h4>Flickagram</h4></a>
```
aaronpanch: You could use a `link_to` here which would clean up the markup.

app/views/tags/index.html.erb
```diff
@@ -1 +1,11 @@
-<%= render partial: 'tagcount', collection: @tags %>
+<div class="col-md-4">
```
aaronpanch: Row?

app/views/layouts/_navbar.html.erb
aaronpanch: Also, why did you choose to use an `h4` here?

mqchau: h4 is to make the logo looks bigger than the default size, I guess the proper way to do it is to assign a class then style it in SCSS file.

app/views/layouts/_footer.html.erb
```diff
@@ -0,0 +1,6 @@
+<footer class="footer bg-inverse">
+  <div class="container">
+    <p class="text-muted footer__text">© AppFolio, Inc. 2016 </p>
```
aaronpanch: Is `.footer_text` used in CSS?

aaronpanch:
Good to merge! 👍

PR #38, opened 2016-08-15
Qc share image

### LINKED ISSUE #10: Share Image
#### As a user I want to share an image with someone via email.

__Story__:

This application is called Image Sharer. Sure, you can share images by adding
them to our web service, but how will people who do not know about this web
service be able to see those images? In phone calls with our customers we have
discovered that many of our users are manually emailing their friends links to
the images. That is awesome! We can make things more efficent for our users by
directly supporting sharing images through email.

__Acceptance criteria__:
- [ ] There is a link to share an image.
- [ ] After clicking on this link, a bootstrap modal appears with a form to
  send the email.
- [ ] The form has fields for the recipient's email and an optional message.
- [ ] The email we send has an HTML part and a text part.
- [ ] The email contains the image, the custom message if provided, and a link
  inviting the recipient to check out our application.
- [ ] The modal closes and a success message is shown if the email was sent
  successfully.
- [ ] The user is alerted of errors if the operation was unsuccessful.

__Note__: This story is large, so it makes sense to break the implementation
down into more manageable pieces (a la elephant carpaccio). Please start by
implementing a basic sharing feature that uses a new page (not a modal), and
open a PR for that. Then, once we've verified the simplified version is
working, refactor it to use a modal, and open a second PR for that. The
[video](https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/basics-of-bootstrap-modals-js-and-ajax)
provides some background and code examples that will be really helpful for the
second PR, so it's worth watching (and following along) before you begin that
part.

__Discussion topics__:
- [ ] How could we develop the front end and back end separately? (And why
  might that be a good idea?)
- [ ] How should we implement a non-CRUD action?
- [ ] What are the different types of errors we should handle?
- [ ] What is AJAX? What libraries can help us add AJAX to our Rails app?
- [ ] How should we go about sending emails from our Rails app?
- [ ] How could we show a preview of the image being shared in the modal?
- [ ] What are the implications of allowing users to send emails from our app?

__Reference material__:
- http://guides.rubyonrails.org/action_mailer_basics.html
- https://elements.heroku.com/addons/sendgrid
- https://devcenter.heroku.com/articles/sendgrid#ruby-rails
- https://github.com/rails/jquery-ujs/wiki
- http://v4-alpha.getbootstrap.com/components/modal
- https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/basics-of-bootstrap-modals-js-and-ajax

__Dependencies__:
- Image Index

mqchau:
Closes #10


rmacklin:
Did you check your review app? Sharing images does not work.

rmacklin:
The first commit in this PR is unnecessarily large. It does too many things at once, which makes it much more likely that a bug will be added. It also makes it more likely that a bug would be overlooked in code review. Additionally, if a bug from that commit did make it to master, the bug couldn't be fixed by simply reverting the commit, because that would revert a bunch of unrelated changes as well. That's not a good property. Please split that commit up into atomic operations (and in general, please _always_ make atomic commits).

Here are some resources about atomic commits:
https://en.wikipedia.org/wiki/Atomic_commit#Atomic_commit_convention
https://medium.com/@fagnerbrack/one-commit-one-change-3d10b10cebbf
https://www.freshconsulting.com/atomic-commits/

mqchau:
So I fixed the environment variables to allow email to work. I'm using Mailgun to send emails and in this dev mode you have to select a list of approved email address to send to. As of now I have quan.chau@appfolio.com added. Do you want me to add your email address richard.macklin@appfolio.com to that too? This way you can choose to send email to your appfolio mail.

mqchau:
Ok I'll work on splitting first commit

rmacklin:
On the topic of atomic operations, I'm wondering why you disregarded the discussion to first get basic sharing working (and open a PR for that), and then refactor to use a modal (and open a second PR for that):
https://github.com/AppFolioOnboarding/image-sharer-mqchau/issues/10#issuecomment-236022854
https://github.com/AppFolioOnboarding/image-sharer-mqchau/issues/10#issuecomment-237429755
https://github.com/AppFolioOnboarding/image-sharer-mqchau/issues/10#issuecomment-237461756
https://github.com/AppFolioOnboarding/image-sharer-mqchau/issues/10#issuecomment-237624612

rmacklin:
> So I fixed the environment variables to allow email to work. I'm using Mailgun to send emails and in this dev mode you have to select a list of approved email address to send to. As of now I have my email address added. Do you want me to add your email address to that too? This way you can choose to send email to your appfolio mail.

This is not an acceptable approach. Any email needs to work without being on a whitelist.

mqchau:
Yeah that was my mistake. Those comments came after I got the modal box working and squashed commits together.

rmacklin:
> Yeah that was my mistake. Those comments came after I got the modal box working and squashed commits together.

That's not true. I looked at your branch at that time and the modal commits had not been squashed together (hence [this comment](https://github.com/AppFolioOnboarding/image-sharer-mqchau/issues/10#issuecomment-237624612)). The commits will still be in your reflog.

mqchau:
I'm trying to fix my mistake here and have splitted the first commit into 3 sub commits that makes sense to do:
- Add mailer
- Add controller
- Add modal box

I think this is more comprehensive than the original history. I agree this way is more cumbersome than if I had just followed your instruction. This is is a lesson to learn.

app/views/image_mailer/image_email.html.erb
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
```
rmacklin: This generic wrapping markup belongs in a mailer _layout_, not in the individual mailer _template_

app/views/image_mailer/image_email.html.erb
```diff
@@ -0,0 +1,10 @@
...
+    <%= @message %><br/>
```
rmacklin: What will happen if they didn't provide a message?

app/views/image_mailer/image_email.text.erb
```diff
@@ -0,0 +1,3 @@
+<%= @message %>
```
rmacklin: What will happen if they didn't provide a message?

app/views/image_mailer/image_email.html.erb
```diff
@@ -0,0 +1,10 @@
...
+    <%= image_tag(@image_link, style: 'max-width: 100px;') %>
```
rmacklin: Why are you limiting this to only 100px? That's quite narrow.

config/environments/development.rb
```diff
@@ -28,7 +28,8 @@
 
   # Don't care if the mailer can't send.
   config.action_mailer.raise_delivery_errors = false
-
+  config.action_mailer.delivery_method = :test
+  config.action_mailer.perform_deliveries = false
```
rmacklin: Are you sure this isn't the development default?

config/environments/development.rb
```diff
@@ -28,7 +28,8 @@
 
   # Don't care if the mailer can't send.
   config.action_mailer.raise_delivery_errors = false
-
+  config.action_mailer.delivery_method = :test
```
rmacklin: Does this need to be set?

config/environments/production.rb
```diff
@@ -56,6 +56,15 @@
...
   config.action_mailer.perform_caching = false
+  config.action_mailer.delivery_method = :smtp
+  ActionMailer::Base.smtp_settings = {
+    port: 587,
+    address: 'smtp.mailgun.org',
+    domain: ENV['domain'],
+    user_name: ENV['username'],
+    password: ENV['password'],
+    authentication: :plain
+  }
```
rmacklin: See https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#issuecomment-239900228

test/mailers/image_mailer_test.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    assert_emails 1 do
+      email.deliver_now
+    end
+
+    assert_equal ['sharing@flickagram.com'], email.from
+    assert_equal ['me@example.com'], email.to
+    assert_equal 'Somebody shared an image with you', email.subject
+    assert_match(/hi there/, email.parts[0].body.to_s)
+    assert_match(%r{http:\/\/www\.yahoo\.com}, email.parts[0].body.to_s)
```
rmacklin: If you use `%r` then you don't need to escape the slashes.

test/mailers/image_mailer_test.rb
```diff
@@ -0,0 +1,19 @@
...
+    assert_match(/hi there/, email.parts[1].body.to_s)
+    assert_match(%r{http:\/\/www\.yahoo\.com}, email.parts[1].body.to_s)
+  end
```
rmacklin: It's not clear what `parts[0]` and `parts[1]` are. Additionally, there's no need to use regex matching for simple string inclusion assertions. You can also remove some repetition from this. An improved version would be:
```ruby
    [:text_part, :html_part].each do |part|
      email_body = email.send(part).body.to_s
      assert_includes email_body, 'http://www.yahoo.com'
      assert_not_includes email_body, 'hi there'
    end
```
Moreover, you probably want to assert that the HTML part actually contains a proper image tag.

test/mailers/image_mailer_test.rb
```diff
@@ -0,0 +1,19 @@
...
+end
```
rmacklin: As indicated by earlier comments, you should write a test for the case when no message was provided.

app/controllers/images_controller.rb
```diff
@@ -48,6 +48,19 @@ def update
     end
   end
 
+  def email
```
rmacklin: Why not call this action `share`?

config/routes.rb
```diff
@@ -2,4 +2,7 @@
   resources :images
   root 'images#index'
   resources :tags, only: [:index]
+  scope '/images/:id' do
+    post 'share', to: 'images#email'
+  end
```
rmacklin: You should implement this route as a `member` route of the `:images` resource. See the rails routing documentation for more information.

test/controllers/images_controller_test.rb
```diff
@@ -192,4 +192,19 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_select '.has-danger'
     assert_select "form input[name='image[tag_list]'][value=?]", ''
   end
+
+  test 'should send email' do
+    image = Image.create!(link: @sample_img_link, tag_list: @sample_img_tags)
+
+    assert_difference 'ActionMailer::Base.deliveries.size', +1 do
```
rmacklin: It's idiomatic to use `1` rather than `+1` in the `assert_difference` here.

test/controllers/images_controller_test.rb
```diff
@@ -192,4 +192,19 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+      post share_path(image), xhr: true, params: { share: { email_address: 'abc@cde.com', message: 'hello' } }
```
rmacklin: I don't think `xhr: true` does what you think it does. You should use the `xhr` method instead
(`xhr :post #...`)

app/controllers/images_controller.rb
```diff
@@ -48,6 +48,19 @@ def update
...
+    email_address = share_params[:email_address]
+    message = share_params[:message]
+    respond_to do |format|
+      if EmailValidator.valid?(email_address)
+        ImageMailer.image_email(email_address, @image.link, message).deliver_now
+        format.json { render json: {} }
```
rmacklin: You don't need to respond with an empty JSON object because you don't need a response body at all. You can simply `head :ok` here (or `head :no_content`)

test/controllers/images_controller_test.rb
```diff
@@ -192,4 +192,19 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+      assert_response :success
+      assert_equal({}, ActiveSupport::JSON.decode(response.body))
+    end
+
+    last_email = ActionMailer::Base.deliveries.last
+
+    assert_equal 'Somebody shared an image with you', last_email.subject
+    assert_equal 'abc@cde.com', last_email.to[0]
```
rmacklin: What else should this test assert?

test/controllers/images_controller_test.rb
```diff
@@ -192,4 +192,19 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  end
```
rmacklin: What other controller tests are necessary?

app/assets/javascripts/image_mail.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Created by appfolio on 8/3/16.
+ */
```
rmacklin: Please remove these comments. They do not add any value.

app/views/image_mailer/image_email.html.erb
mqchau: Then this field is going to be empty. I tried omitting "message" field in post params in controller test and it worked out fine.

app/views/image_mailer/image_email.html.erb
mqchau: Increased to 400px;

test/mailers/image_mailer_test.rb
mqchau: I later modify this part to this:

```ruby
    Array(0..1).each do |part_idx|
      assert_match 'hi there', email.parts[part_idx].body.to_s
      assert_match 'http://www.yahoo.com', email.parts[part_idx].body.to_s
      assert_match 'http://localhost:3000/images/1', email.parts[part_idx].body.to_s
    end
```

Is this ok too?

app/controllers/images_controller.rb
mqchau: Originally action share is to get the separate share page. I guess now this action could be share then.

rmacklin:
The separated commits are an improvement but they're still composed of multiple atomic operations. For example, email configuration setup should be changed independently from creating a new mailer. Adding an `email_validator` gem is also an independent change.

app/controllers/images_controller.rb
bboe: Is `head :no_content` necessary in rails 5?

http://blog.bigbinary.com/2016/07/03/controller-actions-default-no-content-in-rails-5-if-template-is-missing.html

app/assets/javascripts/image_mail.js
```diff
@@ -0,0 +1,20 @@
...
+$(document).ready(function() {
```
rmacklin: Waiting for document ready is an antipattern. More generally, these front end changes need a lot of work. This is why I wanted you to spit up the story into two PRs. Since you didn't do that, you'll need to rewrite a lot of this. Please watch this video https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/basics-of-bootstrap-modals-js-and-ajax and implement your javascript in a similar fashion.

app/views/image_mailer/image_email.html.erb
rmacklin: But you'll have an extra newline. We want to avoid that.

test/mailers/image_mailer_test.rb
rmacklin: No, because it's still unclear what `parts[0]` and `parts[1]` are. Using `[:text_part, :html_part]` makes it clear. Additionally, you're repeating `email.parts[part_idx].body.to_s` several times unnecessarily.

test/controllers/images_controller_test.rb
mqchau: Do you mean to add controller test to invalid image id? Like /images/-1/share or /images/999999/share?

app/controllers/images_controller.rb
rmacklin: I suppose not, but I'd prefer being explicit (and either way, this still shouldn't respond with an empty JSON object)

test/controllers/images_controller_test.rb
rmacklin: Yes

test/controllers/images_controller_test.rb
mqchau: Do you mean to check the body of the email?

test/controllers/images_controller_test.rb
mqchau: I tried what you suggest and this is the warning I got:

DEPRECATION WARNING: ActionDispatch::IntegrationTest HTTP request methods will accept only
the following keyword arguments in future Rails versions:
params, headers, env, xhr, as

Examples:

get '/profile',
  params: { id: 1 },
  headers: { 'X-Extra-Header' => '123' },
  env: { 'action_dispatch.custom' => 'custom' },
  xhr: true,
  as: :json
 (called from block (2 levels) in <class:ImagesControllerTest> at /Users/appfolio/src/image-sharer-mqchau/test/controllers/images_controller_test.rb:200)
DEPRECATION WARNING: xhr and xml_http_request methods are deprecated in favor of
`get "/posts", xhr: true` and `post "/posts/1", xhr: true`.
 (called from block (2 levels) in <class:ImagesControllerTest> at /Users/appfolio/src/image-sharer-mqchau/test/controllers/images_controller_test.rb:200)

rmacklin: Oh I didn't realize this was using ActionDispatch::IntegrationTest because it's a Rails 5 app. Ok, in that case, `xhr: true` looks correct.

app/assets/javascripts/image_mail.js
mqchau: Yeah I've admittedly been abusing $(document).ready for quite a while. I have pushed the javascript code to the bottom of body and include the image_mail.js in similar fashion as the video said.

mqchau: I notice that you're using ES6 in the video. Can I use plain simple javascript and jQuery like right now? It works as far as I can tell.

rmacklin: Your question is a bit confusing. ES6 is just the newest javascript version. jQuery, and all existing javascript, still works in ES6.

Are you're asking if you should be writing ES6 classes? If so, the answer is yes. The javascript for this app is pretty simple, but we want you to get in the habit of writing clean, modular, object-oriented javascript just like you're used to writing in Ruby. In the past, it's been common to treat JS as a "second class citizen", but with the rise of clientside development, that's not a good approach. The code from the video should get you most of the way there (you can view the commits here: https://github.com/AppFolioOnboarding/image-sharer-js-session/compare/01ab975...0562a08).

rmacklin:
This has merge conflicts with master. Additionally, the history is messy. Please rebase and rework the history into atomic, linear commits. When that's done, reply here and I will give another review. Thanks!

mqchau:
I have rewritten history to make it cleaner. And the branch is also up to date with master.

test/controllers/images_controller_test.rb
```diff
@@ -222,4 +222,17 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  test 'should not send email with invalid address' do
+    image = Image.create!(link: @sample_img_link, tag_list: @sample_img_tags)
+
+    assert_no_difference 'ActionMailer::Base.deliveries.size' do
+      post share_image_path(image), xhr: true, params:
+        { share: { email_address: 'abc@cdecom', message: 'hello' } }
+    end
+    assert_response :unprocessable_entity
+    error_obj = ActiveSupport::JSON.decode(response.body)
+    assert error_obj.key?('error')
```
rmacklin: This assertion is redundant since it's implicitly asserted in the subsequent assertion.

app/controllers/images_controller.rb
```diff
@@ -51,8 +51,14 @@ def update
...
     message = share_params[:message]
-    ImageMailer.image_email(email_address, @image.link, image_url(@image), message).deliver_now
-    head :ok
+    if EmailValidator.valid?(email_address)
+      ImageMailer.image_email(email_address, @image.link, image_url(@image), message).deliver_now
+      head :ok
+    else
+      respond_to do |format|
+        format.json { render json: { error: 'Invalid email' }, status: :unprocessable_entity }
+      end
```
rmacklin: We can have the contract of this action be to respond with JSON regardless (i.e. just `render json: { error: 'Invalid email' }, status: :unprocessable_entity` with no `respond_to` blocks)

mqchau:
Hi Richard, I splitted that commit as you said.

app/views/images/_imagebox.html.erb
```diff
@@ -7,6 +7,14 @@
   </div>
   <div class="card-block">
     <%= link_to 'Edit', edit_image_path(image), class: 'btn btn-sm btn-success' %>
+    <button
+      type="button"
+      js-image-share="<%= url_for share_image_path(image) %>"
```
rmacklin: Why are you using `url_for`?

app/views/images/_sharemodal.html.erb
```diff
@@ -0,0 +1,21 @@
+<div id="share_image_modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="share_image_modal" aria-hidden="true">
```
rmacklin: The `aria-labelledby` attribute is incorrect. That should be the id of the title element.

app/views/images/index.html.erb
```diff
@@ -1,4 +1,5 @@
 <h1 class="h1">Images <%= "tagged #{@tag}" if @tag.present? %></h1>
+<%= render partial: 'sharemodal' %>
```
rmacklin: This should be `share_modal`. Also, you should move this partial to the end; since it's hidden it makes sense to render it last.

app/views/images/show.html.erb
```diff
@@ -1,3 +1,4 @@
+<%= render partial: 'sharemodal' %>
```
rmacklin: Same as above

app/assets/javascripts/images/util/share_modal.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  setSharePath(path){
+    this.$share_path = path;
+  }
+
+  attachEventHandlers(){
+    this.$modal.on('submit', event => {
+      if (this.$share_path.length === null) throw new Error('invalid share path');
+      $.post(this.$share_path,
```
rmacklin: I like that you did this with a manual ajax request (it's good experience); however, I'm going to ask you to refactor this to use [jquery_ujs](https://github.com/rails/jquery-ujs) as linked in the issue and mentioned in the video so that you get familiar with it (because it's used very widely in APM).

app/assets/javascripts/images/util/share_modal.js
```diff
@@ -0,0 +1,36 @@
...
+        {share: {email_address: this.$email_address_field.val(), message: this.$message_field.val()}})
+        .done( () => {
+          alert('Email sent successfully');
+          this.$modal.modal('hide');
+        })
+        .fail( () => {
+          alert('Error in sending email');
```
rmacklin: We can and should handle different error types differently. A browser alert is appropriate for unexpected errors, but for a not found error or a validation error we can give a better experience.

For a not found error, a good experience would be to set a flash message on the server and then redirect the user to the index page on the client. That way, they'll see a nicely formatted error message and the image they had tried to share (which had subsequently been deleted) will no longer be there, so they won't be able to repeat the same error.

For a validation error, you should rerender your simple_form with an inline error message to give a consistent experience with how validation errors are handled throughout the rest of the application. This will require some refactoring of your server side code, which will give you the opportunity to implement a form object (which is another thing we want you to practice in this story).

app/views/images/index.html.erb
```diff
@@ -11,3 +11,7 @@
     </div>
   <% end %>
 </div>
+
+<% content_for :javascript do %>
+    <%= javascript_include_tag 'images/index/index' %>
```
rmacklin: Minor, but the indentation is too far here.

test/flows/image_flow_test.rb
```diff
@@ -62,4 +62,27 @@ class ImageFlowTest < FlowTestCase
...
+    fill_in('share[email_address]', with: 'abc@cde.com')
+    fill_in('share[message]', with: 'hi')
+    find('input[type="submit"]').click
+
+    alert = page.driver.browser.switch_to.alert
+    assert_equal 'Email sent successfully', alert.text
+    alert.accept
+
+    page.assert_no_selector('div#share_image_modal.modal.fade.in')
+  end
```
rmacklin: The error paths should be tested as well.

app/assets/javascripts/images/util/image_box.js
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  attachEventHandlers() {
+    this.$share_button.click(event => {
+      this.$share_modal.setSharePath(this.$share_button.attr('js-image-share'));
+    });
+  }
+}
+
+export default ImageBox;
```
rmacklin: This class shouldn't be necessary if you simply take advantage of the bootstrap modal events. See the video for guidance with this.

app/assets/javascripts/images/index/main.js
```diff
@@ -0,0 +1,11 @@
+import $ from 'jquery';
+import ShareModal from '../util/share_modal';
```
rmacklin: I don't think this should be in a `util` directory. It's specific to the images resource, and is not a general utility, so it's fine to just put it in `javascripts/images`

app/views/images/_share_modal.html.erb
```diff
@@ -0,0 +1,21 @@
+<div id="share_image_modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="gridModalLabel" aria-hidden="true">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title" id="gridModalLabel">Share Image</h4>
```
rmacklin: This `id` is from the boostrap modal examples and does not make sense in this context. `share_modal_title` would be an appropriate id.

app/views/layouts/application.html.erb
```diff
@@ -5,7 +5,6 @@
     <%= csrf_meta_tags %>
 
     <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
```
rmacklin: The example code may have missed this, but you should remove the `data-turbolinks-track` attribute from this tag as well.

rmacklin:
Please address this https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#discussion_r74822700. I know it's simple, but we want you to get experience adding conditional logic to a mailer template.

app/views/layouts/application.html.erb
```diff
@@ -18,5 +18,6 @@
     <%= render partial: 'layouts/footer' %>
 
     <%= javascript_include_tag 'application' %>
+    <%= yield :javascript %>
```
rmacklin: This change could have been made independently prior to the rest of this commit. [Remember](https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#issuecomment-239897820), smaller, more focused commits are better for everyone.

app/assets/javascripts/images/util/share_modal.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  attachEventHandlers(){
+    this.$form.on('ajax:success', () => {
+      alert('Email sent successfully');
+      this.$modal.modal('hide');
+    });
+
+    this.$form.on('ajax:error', () => {
+      alert('Error in sending email');
```
rmacklin: See https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#discussion_r75037304

rmacklin:
Suppose I am interrupted briefly after clicking Share. When I come back to this screen:

screen shot 2016-08-16 at 4 28 05 pm

how can I know which image I am sharing? A better user experience to avoid this confusion would be to include a preview of the image inside the modal.

rmacklin:
Please fix this bug:
sharemodalbug

rmacklin:
This PR does not contain the necessary changes to set up the SendGrid addon for all future review apps.

rmacklin:
Please address the review comments in NEW commits. After I review again, we can squash things where it's appropriate. But until then, squashing will make my job as a reviewer more time consuming because I would have to look at all the changes I've already seen again.

app/views/layouts/application.html.erb
mqchau: My thought was that this line wasn't necessary until this commit because this commit add page-specific javascript. Before this commit all pages have same set of javascript included in application.js

rmacklin: That's true that it wasn't necessary before that commit. But it's still its own change, independent from any other changes (the rest of this commit depends on this change, so this change must come before, but this change itself is independent).

mqchau: Ok I'll refactor this after we review other things and then rearrange commits.

mqchau:
You can check my commits starting at "improve on error handling of share form" when you're available. Thank you.

app/assets/javascripts/images/util/share_modal.js
```diff
@@ -17,8 +17,12 @@ class ShareModal {
       this.$modal.modal('hide');
     });
 
-    this.$form.on('ajax:error', () => {
-      alert('Error in sending email');
+    this.$form.on('ajax:error', (event, xhr, status) => {
+      if (xhr.status === 404) {
+        window.location.href = '/'; //one level up
```
rmacklin: This comment is confusing, and IMO unnecessary

mqchau: Agree. Deleted that comment in a subsequent commit.

app/controllers/images_controller.rb
```diff
@@ -49,18 +49,26 @@ def update
...
 
   def share
-    email_address = share_params[:email_address]
-    message = share_params[:message]
-    if EmailValidator.valid?(email_address)
-      ImageMailer.image_email(email_address, @image.link, image_url(@image), message).deliver_now
+    @image_share_form = ImageShareForm.new(share_params)
+    if @image_share_form.valid?
+      ImageMailer
+        .image_email(share_params[:email_address], @image.link, image_url(@image), share_params[:message])
```
rmacklin: It would be better to get these values out of the form object (which could potentially decorate them), rather than from the raw `params`.

app/assets/javascripts/images/share_modal.js
```diff
@@ -5,28 +5,34 @@ class ShareModal {
...
       }
     });
 
     this.$form.on('ajax:complete', () => {
       this.$submit_button.prop('disabled', false);
+      this.$email_address.val('');
+      this.$message.val('');
+    });
+
+    // Change image share path in form each time the like button is clicked
```
rmacklin: I think you mean the share button. But, I don't think either of the comments added in this file are necessary; they don't add anything that's not already clear from the code.

app/views/image_mailer/image_email.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= @message %><br/>
+<%= @message + "<br/>" unless @message.to_s == '' %>
```
rmacklin: Did you test this change? By default, rails will HTML-escape strings that aren't explicitly marked as safe. Go ahead and share an image with a message to yourself and you'll see the string `<br/>` is inclued in the email.

To solve this, we can change this to:

```erb
<% if @message.present? %>
  <%= @message>
  <br>
<% end %>
```
(by the way, `<br>` tags don't need to be closed)

You could even customize it a tad more, like:
```erb
<% if @message.present? %>
  Your friend said:
  <br>
  <i><%= @message></i>
  <br>
<% end %>
```

app.json
```diff
@@ -21,7 +21,8 @@
     }
   },
   "addons": [
-    "heroku-postgresql"
+    "heroku-postgresql",
+    "heroku-sendgrid"
```
rmacklin: This is correct, but you'll also want to make it inherit the SendGrid ENV variables, so that you don't have to configure those for each subsequent review app.

app/assets/javascripts/images/share_modal.js
```diff
@@ -3,10 +3,11 @@ import $ from 'jquery';
 class ShareModal {
   constructor(modalSelector){
     this.$modal = $(modalSelector);
-    this.$submit_button = this.$modal.find('#image_mail_submit');
+    this.$image_preview = this.$modal.find('.js-image-share-preview');
     this.$form = this.$modal.find('form');
     this.$email_address = this.$form.find('#image_share_form_email_address');
     this.$message = this.$form.find('#image_share_form_message');
+    this.$submit_button = this.$form.find('#image_mail_submit');
```
rmacklin: At first I was confused why this was being added. Now I see that it was just moved. It's best to avoid making changes like that, unless it's part of its own dedicated commit, to reduce confusion.

app.json
mqchau: So to add  
"env": {
  "SENDGRID_USERNAME": {
     "required": true
  }
}

to app.json?

app/assets/javascripts/images/share_modal.js
mqchau: Well it was moved because I think it belongs to $form, not the $modal, so I put it after this.$form is defined.

app.json
rmacklin: Yes, and also add the SENDGRID_PASSWORD variable. BTW, you should surround code blocks in ``` so that GitHub will format them correctly.

rmacklin:
You partially fixed the bug I mentioned, but it's still there:

sharemodalbug2

Can you think of a more robust fix?

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,41 @@
...
+    this.$form.on('ajax:error', (event, xhr) => {
+      if (xhr.status === 404) {
+        window.location.href = '/';
+      } else {
+        this.$form.html(xhr.responseJSON.form);
+      }
+    });
+
+    this.$form.on('ajax:complete', () => {
+      this.$submit_button.prop('disabled', false);
```
rmacklin: Are you sure this is necessary? jquery_ujs should handle the re-enabling for you.

app/views/images/_imagebox.html.erb
```diff
@@ -7,6 +7,14 @@
   </div>
   <div class="card-block">
     <%= link_to 'Edit', edit_image_path(image), class: 'btn btn-sm btn-success' %>
+    <button
+      type="button"
+      js-image-share="<%= share_image_path(image) %>"
```
rmacklin: `js-image-share` is not a valid attribute for `button` tags, so this is invalid HTML. It will work in most browsers, but the proper way to declare custom attributes for any element is to prefix them with `data-` (these are unsurprisingly called "data attributes"). So, this would be `data-image-share`, though a more descriptive name would be `data-share-path`.

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,37 @@
...
+      } else {
+        this.$form.html(xhr.responseJSON.form);
+      }
+    });
+
+    this.$modal.on('show.bs.modal', e => {
+      this.$email_address.val('');
+      this.$message.val('');
+      this.$form.attr('action', $(e.relatedTarget).attr('js-image-share'));
+      this.$image_preview.attr('src', $(e.relatedTarget).closest('.js-image-box').find('img').first().attr('src'));
```
rmacklin: This works, but a cleaner approach would just be to use another [data attribute](https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#discussion_r75224440) for the image url on each share button.

rmacklin: If you change this to using two data attributes, you can clean this up nicely by extracting a `setupForm` instance method:
```js
class ShareModal {
  // ...

  setupForm(sharePath, imageUrl) {
    this.$form.attr('action', sharePath);
    this.$imagePreview.attr('src', imageUrl);
  }
}
```
and calling that here.

rmacklin: If you use data attributes, you can use the `dataset` property to simplify this so that it would be called like:
```js
this.setupForm(e.relatedTarget.dataset.sharePath, e.relatedTarget.imageUrl)
```
And then you can actually take advantage of ES6 destructuring to make it even cleaner. First you'd change the method signature to:
```js
setupForm({sharePath, imageUrl}) {
  // ...
}
```
and then you'd change the invocation to just:
```js
this.setupForm(e.relatedTarget.dataset);
```
This kind of mimics a React-like thinking where the button's dataset act as the component state for the share form.

rmacklin: By the way, that was discussed in the video. If you didn't watch the whole thing, I'd highly recommend you do so.

app/assets/javascripts/images/share_modal.js
rmacklin: Yeah, it makes sense why you moved it. But that explanation is exactly the kind of thing that should be included in the commit message of an atomic commit for just that change - that way, I wouldn't even have had to ask you why you moved it.

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,32 @@
...
+      if (xhr.status === 404) {
+        window.location.href = '/';
+      } else {
+        this.$form.html(xhr.responseJSON.form);
+      }
+    });
+
+    this.$modal.on('show.bs.modal', e => {
+      this.$form.attr('action', $(e.relatedTarget).attr('data-share-path'));
+      this.$image_preview.attr('src', $(e.relatedTarget).closest('.js-image-box').find('img').first().attr('src'));
```
rmacklin: See https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#discussion_r75224787. That comment thread now appears as "on an outdated diff" because you are pushing while I am in the middle of adding my review comments. Typically we try to avoid that.

test/flows/image_flow_test.rb
rmacklin: There is still another error path that is not tested here.

test/flows/image_flow_test.rb
```diff
@@ -62,4 +62,35 @@ class ImageFlowTest < FlowTestCase
     page.assert_selector("img[src='#{sample_img_link1}']")
     page.assert_selector("img[src='#{sample_img_link2}']")
   end
+
```
rmacklin: You should write a flow test to verify that the bug (see [this](https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#issuecomment-240271374) and [this](https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#issuecomment-240576263)) I found has been fixed.

app/assets/javascripts/images/share_modal.js
```diff
@@ -3,8 +3,8 @@ import $ from 'jquery';
 class ShareModal {
   constructor(modalSelector){
     this.$modal = $(modalSelector);
-    this.$image_preview = this.$modal.find('.js-image-share-preview');
     this.$form = this.$modal.find('form');
+    this.$clean_form = this.$form.html();
```
rmacklin: Our convention is to use camelCase rather than snake_case for javascript identifiers.

rmacklin:
Please see https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#discussion_r75224787. That comment thread now appears as "on an outdated diff" because you are pushing while I am in the middle of adding my review comments. Typically we try to avoid that. (You can still make the changes locally, and then push them once I am done making my comments.)

rmacklin:
You should write a flow test to verify that the bug (see [this](https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#issuecomment-240271374) and [this](https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#issuecomment-240576263)) I found has been fixed.

app/assets/javascripts/images/share_modal.js
mqchau: New JS method doesn't need to define any of those this.$email_address, $message, $submit_button. So I guess this comment is done.

mqchau:
I think that we're quite close to finishing up this story. Do you have time today so we can work together and nail this story?

app/assets/javascripts/images/share_modal.js
```diff
@@ -22,11 +22,10 @@ class ShareModal {
...
     this.$modal.on('show.bs.modal', e => {
-      this.$form.html(this.$clean_form);
-      this.$form.attr('action', $(e.relatedTarget).attr('data-share-path'));
-      this.$form.find('.js-image-share-preview').attr( 'src',
-        $(e.relatedTarget).closest('.js-image-box').find('img').first().attr('src')
-      );
+      // TODO: test the form is cleaned after each load
+      this.$form.html(this.$cleanForm);
+      this.$form.attr('action', e.relatedTarget.dataset.sharePath);
+      this.$form.find('.js-image-share-preview').attr('src', e.relatedTarget.dataset.imageLink);
```
rmacklin: It would be good to clean this up further as described in https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#discussion_r75225492

test/flows/image_flow_test.rb
```diff
@@ -92,5 +95,10 @@ class ImageFlowTest < FlowTestCase
     alert.accept
 
     page.assert_no_selector('div#share_image_modal.modal.fade.in')
+
+    click_button 'Share'
+    assert_equal Image.last.link, find('.js-image-share-preview')[:src]
+    assert_equal '', find_field('image_share_form[email_address]').value
+    assert_equal '', find_field('image_share_form[message]').value
```
rmacklin: This test doesn't fully cover the bug I found. You need to have two images, share one with an invalid email, then share the other one and verify that a) the preview image is for the correct image, b) the fields have been emptied, c) the validation error is no longer displaying. You can do this in a separate test case rather than shoving everything into this test case

test/flows/image_flow_test.rb
```diff
@@ -92,5 +95,10 @@ class ImageFlowTest < FlowTestCase
...
   end
```
rmacklin: There's still an untested code path that should be tested here. Any ideas?

app/assets/javascripts/images/share_modal.js
```diff
@@ -22,11 +22,15 @@ class ShareModal {
     });
 
     this.$modal.on('show.bs.modal', e => {
-      this.$form.html(this.$cleanForm);
-      this.$form.attr('action', e.relatedTarget.dataset.sharePath);
-      this.$form.find('.js-image-share-preview').attr('src', e.relatedTarget.dataset.imageLink);
+      this.setupForm(e.relatedTarget.dataset.sharePath, e.relatedTarget.dataset.imageLink);
```
rmacklin: This is good, but using ES6 destructuring (as described in the [comment](https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#discussion_r75225492)) would be better.

test/flows/image_flow_test.rb
```diff
@@ -62,4 +62,43 @@ class ImageFlowTest < FlowTestCase
...
+    assert_equal 'Email sent successfully', alert.text
+    alert.accept
+
+    page.assert_no_selector('div#share_image_modal.modal.fade.in')
+
+    click_button 'Share'
+    assert_equal Image.last.link, find('.js-image-share-preview')[:src]
+    assert_equal '', find_field('image_share_form[email_address]').value
+    assert_equal '', find_field('image_share_form[message]').value
+  end
```
rmacklin: There's still an untested code path that should be tested here. Any ideas?

test/flows/image_flow_test.rb
rmacklin: You should create a new flow test file for the `ShareImageTest` cases. Also, this test should be renamed to `ImagesCrudTest` since it's for testing the image crud operations.

mqchau:
For some reason the flow test fails on my computer but pass on CircleCI. At this line in the end of image_share_test.rb:96

```ruby
    page.find('.close').click
    assert_no_selector 'div#share_image_modal.modal.fade.in'
```

I think it's a browser issue or race condition. Do you have any suggestion to fix it?

rmacklin:
The modal `.fade.in` classes are suspicious. It seems you still haven't converted your flow tests to use page objects. There are built in methods in ae_page_objects that handle testing modal interactions non-flakily.

mqchau:
The problem is that it checks for the element too quickly before it disappears. I agree this is flaky and noted a TODO to implement this check in the next issue to use PageObject.

mqchau:
Hi Richard, do you have any other comments on this or should I rearrange the commits in clean order?

app/assets/javascripts/images/index/main.js
```diff
@@ -0,0 +1,6 @@
+import $ from 'jquery';
```
rmacklin: This is not used

app/assets/javascripts/images/show/main.js
```diff
@@ -0,0 +1,5 @@
+import ShareModal from '../share_modal';
+
+const shareModal = new ShareModal('#share_image_modal');
+
+shareModal.attachEventHandlers();
```
rmacklin: This is exactly the same as `index/main.js`. Always try to avoid duplication ([DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)). If another developer were to work on this project, they may not know they need to make changes to both files when fixing a bug or adding a new feature. But if there was only one file, there would be no issue. Please DRY this.

app/controllers/images_controller.rb
```diff
@@ -1,6 +1,6 @@
 class ImagesController < ApplicationController
   before_action :set_image, only: [:show, :destroy, :update, :edit]
-
+  before_action :set_image_with_flash, only: [:share]
```
rmacklin: Previously there was an empty line between the `before_action` and `def index`, for readability. Please be careful not to remove empty lines that are there for readability when you are adding new stuff.

app/controllers/images_controller.rb
```diff
@@ -48,8 +48,27 @@ def update
...
+    else
+      render json: { form: render_to_string(partial: '/images/share_form') }, status: :unprocessable_entity
+    end
+  end
+
   private
 
+  def set_image_with_flash
+    set_image
+  rescue ActiveRecord::RecordNotFound
```
rmacklin: Generally exceptions should be reserved for _exceptional_ behavior, not used for control flow. In this case you can use `Image.find_by(id: params[:id])` which will return either an `Image` instance or `nil` (rather than raising an exception), and then you won't have to use the exception for control flow.

app/views/images/_share_form.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= simple_form_for @image_share_form || ImageShareForm.new, url: '', remote: true, html: { id: 'image_mail_form', novalidate: false } do |f| %>
+    <div class="modal-body row">
```
rmacklin: Please use 2 spaces for indentation

rmacklin: Please use 2 spaces for indentation here and elsewhere in the PR

test/controllers/images_controller_test.rb
```diff
@@ -192,4 +192,47 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  test 'should not send email with invalid address' do
+    image = Image.create!(link: @sample_img_link, tag_list: @sample_img_tags)
+
+    assert_no_difference 'ActionMailer::Base.deliveries.size' do
+      post share_image_path(image), xhr: true, params:
+        { image_share_form: { email_address: 'abc@cdecom', message: 'hello' } }
+    end
+    assert_response :unprocessable_entity
+    error_obj = ActiveSupport::JSON.decode(response.body)
+    assert error_obj.key?('form')
```
rmacklin: What else can you assert about the value of `form`?

test/flows/image_share_test.rb
```diff
@@ -0,0 +1,100 @@
...
+
+class ImageShareTest < FlowTestCase
+  test 'add new image and share it' do
+    sample_img_link = 'http://i.imgur.com/PVrX7wZ.jpg'
+    tag_name = 'testtag1'
+
+    visit(new_image_path)
+    fill_in('image[link]', with: sample_img_link)
+    fill_in('image[tag_list]', with: tag_name)
+    find('input[type="submit"]').click
```
rmacklin: You don't need to add images throw the UI anywhere that you're not explicitly testing that new image form. You should just create `Image` records in the database at the beginning, which will save a lot of time in your tests.

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,37 @@
...
+      this.$modal.modal('hide');
+    });
+
+    this.$form.on('ajax:error', (event, xhr) => {
+      if (xhr.status === 404) {
+        window.location.href = '/';
+      } else {
+        this.$form.html(xhr.responseJSON.form);
+      }
+    });
```
rmacklin: What about other error responses?

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,37 @@
...
+    this.$form.on('ajax:success', () => {
+      alert('Email sent successfully');
+      this.$modal.modal('hide');
+    });
+
+    this.$form.on('ajax:error', (event, xhr) => {
+      if (xhr.status === 404) {
+        window.location.href = '/';
+      } else {
+        this.$form.html(xhr.responseJSON.form);
```
rmacklin: Is this appropriate for all other error responses?

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,37 @@
...
+class ShareModal {
+  constructor(modalSelector){
+    this.$modal = $(modalSelector);
+    this.$form = this.$modal.find('form');
+    this.$cleanForm = this.$form.html();
+  }
+
+  attachEventHandlers(){
+    this.$form.on('ajax:success', () => {
+      alert('Email sent successfully');
```
rmacklin: This isn't a good use a browser alert from a UX perspective. It does not match how we show success messages in response to all the other user actions in our application. We should apply consistent UX.

config/routes.rb
```diff
@@ -1,5 +1,9 @@
 Rails.application.routes.draw do
-  resources :images
+  resources :images do
+    member do
+      post 'share'
```
rmacklin: This can be a symbol rather than a string.

mqchau:
Hi Richard, I made some major changes to frontend js code to display email success message in the flash messages area. Please review when you can. Thank you.

test/flows/image_share_test.rb
```diff
@@ -1,16 +1,15 @@
...
-    find('input[type="submit"]').click
+  setup do
+    @sample_img_link1 = 'http://i.imgur.com/PVrX7wZ.jpg'
+    @sample_img_link2 = 'http://i.imgur.com/zYn8F.jpg'
+    Image.create!(link: 'http://i.imgur.com/PVrX7wZ.jpg', tag_list: 'testtag1')
+    Image.create!(link: 'http://i.imgur.com/zYn8F.jpg', tag_list: 'testtag1')
+  end
 
-    assert has_current_path?(image_path(Image.last))
+  test 'add new image and share it' do
```
rmacklin: This test name is no longer accurate

test/controllers/images_controller_test.rb
```diff
@@ -233,6 +233,6 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     end
     assert_response :unprocessable_entity
     error_obj = ActiveSupport::JSON.decode(response.body)
-    assert error_obj.key?('form')
+    assert_includes error_obj['form'], 'is invalid'
```
rmacklin: There is still more you can check about the contents of `error_obj['form']`. For example, you'd want to verify that it contains the right action, and that the previous input values were preserved.

app/assets/javascripts/images/share_modal.js
```diff
@@ -5,11 +5,17 @@ class ShareModal {
     this.$modal = $(modalSelector);
     this.$form = this.$modal.find('form');
     this.$cleanForm = this.$form.html();
+    this.$flashMessageArea = $('.main-layout__flash-message');
```
rmacklin: Does this belong as a property of `ShareModal` instances?

app/assets/javascripts/images/share_modal.js
```diff
@@ -5,11 +5,17 @@ class ShareModal {
...
   }
 
   attachEventHandlers(){
     this.$form.on('ajax:success', () => {
-      alert('Email sent successfully');
+      if (this.$flashMessageArea.find('#success').length === 0) {
+        var success_message_area = $('<div>').prop('id', 'success').addClass('alert').addClass('alert-success');
```
rmacklin: Do not use `var` in ES6. Also, remember that identifiers should be camelCase.

There's no reason to construct the message iteratively like this. You can simply do:
```js
const successAlert = `
  <div class="alert alert-success"></div>
`;
```

app/assets/javascripts/images/share_modal.js
```diff
@@ -5,11 +5,17 @@ class ShareModal {
...
+        this.$flashMessageArea.append(success_message_area);
+      }
+      var email_address = this.$form.find('#image_share_form_email_address').val();
```
rmacklin: Same: don't use `var` and use camelCase

app/assets/javascripts/images/share_modal.js
```diff
@@ -1,29 +1,28 @@
...
-    this.$flashMessageArea = $('.main-layout__flash-message');
   }
 
   attachEventHandlers(){
     this.$form.on('ajax:success', () => {
-      if (this.$flashMessageArea.find('#success').length === 0) {
-        var success_message_area = $('<div>').prop('id', 'success').addClass('alert').addClass('alert-success');
-        this.$flashMessageArea.append(success_message_area);
-      }
       var email_address = this.$form.find('#image_share_form_email_address').val();
```
rmacklin: Why did you keep this?

app/assets/javascripts/images/share_modal.js
```diff
@@ -1,29 +1,28 @@
...
 
     this.$form.on('ajax:error', (event, xhr) => {
       if (xhr.status === 404) {
         window.location.href = '/';
-      } else {
+      } else if (xhr.status === 422) {
         this.$form.html(xhr.responseJSON.form);
+      } else {
+        FlashMessage.diplayErrorMessage('Email failed to sent to ' + this.getEmailAddress());
+        this.$modal.modal('hide');
```
rmacklin: In this case I think a browser alert would be appropriate UX because this is a completely unexpected error. I don't think we should hide the modal because it's possible that the only thing they need to do is try again.

app/assets/javascripts/shared/flash_message.js
```diff
@@ -0,0 +1,28 @@
...
+    this.clearMessage();
+    var error_message_area = $('<div>').prop('id', 'error').addClass('alert').addClass('alert-danger');
+    this.$flashMessageArea.append(error_message_area);
+    this.$flashMessageArea.find('#error').html(msg);
+  }
+
+  static clearMessage(){
+    this.$flashMessageArea.empty();
+  }
+}
```
rmacklin: There's not really any reason to use a class with only static methods over just a plain old javascript object.

app/views/images/edit.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <div class="row">
   <div class="col-md-4">
-    <%= image_tag @image.link, class: 'image-box__image' %>
+    <%= image_tag @image.link, class: 'image' %>
```
rmacklin: How is this change related to sharing images?

app.json
```diff
@@ -28,7 +28,7 @@
   },
   "addons": [
     "heroku-postgresql",
-    "heroku-sendgrid"
+    "sendgrid"
```
rmacklin: I think this should be `"sendgrid:starter"` to ensure we use the free "starter" plan. In fact, I think `"heroku-postgresql"` should actually be `"heroku-postgresql:hobby-dev"`. Maybe heroku applies the free plans as the default, but I prefer being explicit in case that changes.

app/assets/javascripts/shared/flash_message.js
```diff
@@ -0,0 +1,28 @@
+import $ from 'jquery';
+
+class FlashMessage {
+  static setup(){
+    this.$flashMessageArea = $('.main-layout__flash-message');
```
rmacklin: We should give this element an explicit id (e.g. `flash_messages`) rather than selecting it using a stylistic class (which could be removed in a style update, and that would break this JS)

app/assets/javascripts/shared/flash_message.js
rmacklin: Taking it further, I think we can simplify the whole thing to a single `setFlashMessage` function:

```js
function setFlashMessage(type, message) {
  document.getElementById('flash_messages').innerHTML =
    `<div class="alert alert-${type} %>">${message}</div>`;
}
```

rmacklin:
https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/38#discussion_r75781099 has not been addressed.

app/assets/javascripts/images/share_modal.js
rmacklin: I see it was removed in a later commit

rmacklin:
You have a syntax error that's breaking both tests and rubocop

app/views/images/_share_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= simple_form_for @image_share_form || ImageShareForm.new, url: '', remote: true, html: { id: 'image_mail_form', novalidate: false } do |f| %>
+<%= simple_form_for @image_share_form || ImageShareForm.new, url: @image_share_form.share_path, remote: true, html: { id: 'image_mail_form', novalidate: false } do |f| %>
```
rmacklin: This change indicates that the flow test was missing test coverage.

app/views/images/edit.html.erb
mqchau: I realize it is not related but this is a bug from previous commit. Should I leave it in a separate PR?

rmacklin:
Superseded by #42

PR #39, opened 2016-08-15
fix flash errror message

mqchau:
Patch #9 
If user wants to show images of a non-existent tag, it will display a flash error message on top. And that flash error message persists for another page when user navigates to somewhere else. This patch will fix that by making that flash message to appear only once then disappear when user navigates somewhere else.


bboe:
👍

shiboying:
👍

PR #40, opened 2016-08-15
add configuration to remove unused tags

mqchau:
Patch #16


bboe:
Should we add a unit test for this behavior? This test will likely need to actually hit the database.

mqchau:
So the test is added, can I merge this PR?

bboe:
👍

PR #42, opened 2016-08-31
Qc share image second attempt

### LINKED ISSUE #10: Share Image

mqchau:
Second attempt to close #10


app/mailers/image_mailer.rb
```diff
@@ -0,0 +1,10 @@
+class ImageMailer < ApplicationMailer
+  default from: 'sharing@flickagram.com'
+
+  def image_email(email_address, image_link, application_image_link, message)
```
rmacklin: Consider using named arguments so that the callers are not forced to pass parameters in this order (i.e. this reduces Connascence of Position to Connascence of Name). It will also make invocations [like this](https://github.com/AppFolioOnboarding/image-sharer-mqchau/blob/542f3b51caf520d667a518b1f959253ae7bc26f2/test/mailers/previews/image_mailer_preview.rb#L3-L8) more self documenting.

test/mailers/image_mailer_test.rb
```diff
@@ -0,0 +1,70 @@
+require 'test_helper'
+
+class ImageMailerTest < ActionMailer::TestCase
+  test 'share image with message' do
+    email = ImageMailer.image_email(
+      'me@example.com',
+      'http://www.yahoo.com',
```
rmacklin: It would improve test readability if you used a URL that resembles the real world usage (e.g. a URL that looks like an image URL, rather than `yahoo.com`)

test/controllers/images_controller_test.rb
```diff
@@ -192,4 +192,34 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+      assert_includes email_body, @sample_img_link
+      assert_includes email_body, 'hello'
+      assert_includes email_body, image_url(image)
+    end
+  end
+
+  test 'should not send email with invalid image id' do
+    assert_no_difference 'ActionMailer::Base.deliveries.size' do
+      assert_raises ActiveRecord::RecordNotFound do
+        post share_image_path(999_999), xhr: true, params:
```
rmacklin: The convention is to use `-1` for nonexistent ids.

app/models/image_share_form.rb
```diff
@@ -0,0 +1,7 @@
+class ImageShareForm
```
rmacklin: This is a form object, not a model, so it doesn't belong in `app/models`

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,44 @@
...
+      }
+    });
+
+    this.$modal.on('show.bs.modal', e => {
+      this.setupForm(e.relatedTarget.dataset);
+    });
+  }
+
+  getEmailAddress(){
+    return this.$form.find('#image_share_form_email_address').val();
```
rmacklin: If your server takes more than a few milliseconds to respond, this could be inaccurate because they could have changed the input value (e.g. by deleting characters). If you want to guarantee accuracy, you should cache the value of the input as soon as the form is submitted/the request is sent.

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  setupForm({sharePath, imageLink}){
+    this.$form.html(this.$cleanForm);
+    this.$form.attr('action', sharePath);
+    this.$form.find('.js-image-share-preview').attr('src', imageLink);
+  }
+}
+
+
```
rmacklin: Minor: extra newline

app/assets/javascripts/images/share_modal_initialize.js
```diff
@@ -0,0 +1,5 @@
+import ShareModal from './share_modal';
+
+const shareModal = new ShareModal('#share_image_modal');
+shareModal.attachEventHandlers();
+
```
rmacklin: Minor: extra newline

app/assets/javascripts/images/show/index.js
```diff
@@ -0,0 +1,2 @@
+require('../share_modal_initialize.js');
+
```
rmacklin: Minor: extra newline

app/assets/javascripts/shared/flash_message.js
```diff
@@ -0,0 +1,13 @@
...
+    `<div id="${type}" class="alert ${alertType} %>">${message}</div>`;
+}
+
+function decodeMesssageTypeToAlert(type){
+  if (type === 'error') return 'alert-danger';
+  else return `alert-${type}`;
+}
+
+export default setFlashMessage;
+
```
rmacklin: Minor: extra newline

app/assets/javascripts/shared/flash_message.js
```diff
@@ -0,0 +1,13 @@
+function setFlashMessage(type, message) {
+  var alertType = decodeMesssageTypeToAlert(type);
+  document.getElementById('flash-messages').innerHTML =
+    `<div id="${type}" class="alert ${alertType} %>">${message}</div>`;
```
rmacklin: I don't think this `id` is necessary

rmacklin: Also, it looks like you've got a stray ` %>` here

app/assets/javascripts/shared/flash_message.js
```diff
@@ -0,0 +1,13 @@
+function setFlashMessage(type, message) {
+  var alertType = decodeMesssageTypeToAlert(type);
+  document.getElementById('flash-messages').innerHTML =
```
rmacklin: Our convention is to use underscores as the word separator for `id`s (in contrast to using hyphens as the word separator in classes)

app/assets/javascripts/shared/flash_message.js
```diff
@@ -0,0 +1,13 @@
...
+    `<div id="${type}" class="alert ${alertType} %>">${message}</div>`;
+}
+
+function decodeMesssageTypeToAlert(type){
+  if (type === 'error') return 'alert-danger';
+  else return `alert-${type}`;
+}
+
+export default setFlashMessage;
```
rmacklin: It's a good practice to make the filename match the name of the default export (i.e. use `set_flash_message.js` rather than `flash_message.js`) since that makes importing it intuitive.

app/views/layouts/_flash_message.html.erb
```diff
@@ -1,5 +1,7 @@
-<% flash.each do |key, value| %>
-  <% unless value.empty? %>
-    <div id="<%= key %>" class="alert <%= flash_to_alert_class(key) %>"><%= value %></div>
+<div class="main-layout__flash-message" id="flash-messages">
```
rmacklin: See https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/42#discussion_r77086743

app/views/images/index.html.erb
```diff
@@ -12,3 +12,7 @@
 </div>
 
 <%= render partial: 'share_modal' %>
+
+<% content_for :javascript do %>
+    <%= javascript_include_tag 'images/index/index' %>
```
rmacklin: Indentation is off here

app/views/images/show.html.erb
```diff
@@ -5,3 +5,7 @@
 </div>
 
 <%= render partial: 'share_modal' %>
+
+<% content_for :javascript do %>
+    <%= javascript_include_tag 'images/show/index' %>
```
rmacklin: Indentation is off here

test/flows/image_share_test.rb
```diff
@@ -0,0 +1,90 @@
...
+    @image1 = Image.create!(link: @sample_img_link1, tag_list: 'testtag1')
+    @image2 = Image.create!(link: @sample_img_link2, tag_list: 'testtag1')
+  end
+
+  test 'share image' do
+    visit image_path(Image.last)
+    click_button 'Share'
+
+    assert_selector 'div#share_image_modal.modal.fade.in'
+    assert_includes find('#new_image_share_form')[:action], share_image_path(Image.last)
```
rmacklin: Why is this an `includes` assertion rather than an equality assertion?

test/controllers/images_controller_test.rb
```diff
@@ -216,11 +216,12 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
 
   test 'should not send email with invalid image id' do
     assert_no_difference 'ActionMailer::Base.deliveries.size' do
-      assert_raises ActiveRecord::RecordNotFound do
-        post share_image_path(999_999), xhr: true, params:
-          { image_share_form: { email_address: 'abc@cde.com', message: 'hello' } }
-      end
+      post share_image_path(999_999), xhr: true, params:
```
rmacklin: https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/42#discussion_r77085428

app/models/image_share_form.rb
```diff
@@ -0,0 +1,7 @@
...
+  include ActiveModel::Model
+
+  attr_accessor :email_address, :image_link, :message, :share_path
```
rmacklin: Why does the form object have an `image_link` and `share_path` attribute? Those are not fields in the share form.

app/controllers/images_controller.rb
```diff
@@ -3,6 +3,7 @@ class ImagesController < ApplicationController
   before_action :set_image_with_flash, only: [:share]
 
   def index
+    @image_share_form = ImageShareForm.new(share_path: '', image_link: '')
```
rmacklin: This change is unnecessary if you structure your form partial appropriately.

app/controllers/images_controller.rb
```diff
@@ -15,6 +16,7 @@ def index
   end
 
   def show
+    @image_share_form = ImageShareForm.new(share_path: '', image_link: '')
```
rmacklin: This change is unnecessary if you structure your form partial appropriately.

app/controllers/images_controller.rb
```diff
@@ -85,4 +88,9 @@ def image_tag_params
   def share_params
     params.require(:image_share_form).permit(:email_address, :message)
   end
+
+  def set_up_image_share_form_for_rendering_with_error
+    @image_share_form.share_path = share_image_path(@image)
+    @image_share_form.image_link = @image.link
+  end
```
rmacklin: This method is unnecessary if you structure your form partial appropriately.

app/views/images/_imagebox.html.erb
```diff
@@ -7,6 +7,14 @@
   </div>
   <div class="card-block">
     <%= link_to 'Edit', edit_image_path(image), class: 'btn btn-sm btn-success' %>
+    <button
+      type="button"
+      js-image-share="<%= share_image_path(image) %>"
```
rmacklin: This is an invalid attribute. Furthermore, it's not used in this commit, so it can be removed

test/flows/image_share_test.rb
mqchau: For some reason this action is the full URL http://www.example.com/<the path>, I also tried to call share_image_url(Image.last) but it didn't work.

app/models/image_share_form.rb
mqchau: This is comment from Sarah Jones when I was pairing with her. The way the share form is structured now it actually contains the image link (preview image) and the share path in action attributes. So I think it's natural to bundle them together.

app/assets/javascripts/shared/flash_message.js
mqchau: The id is from other previous issue. That's how it was structured in the template file and I want to keep things consistent. Also later on the PageObject will use this id to locate the flash messages.

app/models/image_share_form.rb
rmacklin: If I understand what you mean by "the way the form is structured now", it sounds like you're commingling the view with the form, and by extension the idea of what would belong in a view model with what would belong in a form object.

The preview image element is _not_ part of the form. Your HTML is currently structured such that it is a child of the form element, but conceptually it is not associated with the form - the client does not send it's value to the server.

The form action is a different case. It is "part of the form", but it's still not a form attribute, and thus it should not be declared with `attr_accessor` here. In Rails forms that are backed by a model which has corresponding resourceful routes, form actions are inferred automatically. But for all other cases (such as this one, since the share form isn't for a resource), the form action is specified using the `url` option of `form_for` (or `simple_form_for`). Normally this is done by calling the Rails url helpers directly in the view, not supplied by a method on the form object. Regardless, if you did want to supply it as a method on the form object, that method shouldn't be an accessor, because again this is not an attribute of the form.

app/assets/javascripts/shared/flash_message.js
rmacklin: I traced the `id` back to here: https://github.com/AppFolioOnboarding/image-sharer-mqchau/commit/b69938ac7518ec3024f461f23ab1c69e3058c6bb#diff-9599427925097c3c66f26ac1e0de5cadR12

That is from PR #23 which you merged yourself without code review. The id was never necessary up to this point, and it still isn't.

The fact that you're using it in the page object branch does not make it necessary for this branch (and it's also not necessary to complete the page object branch)

mqchau: It is used in test/flows/image_flow_test.rb . And that was there from previous issues.

rmacklin: It doesn't need to be used there. Anywhere that is selecting by these ids can be replaced equivalently to select the alert of the appropriate `type` inside the element with the `flash_messages` id. The reason I'm mentioning this is that people sometimes have a tendency to add ids to things all the time even when they don't need them, and we want to avoid that. Since you still have not completed the page objects refactor, I won't make you change this now, but you should do so before that branch is merged.

app/models/image_share_form.rb
mqchau: Ok I just added 3 separate commits to address this. So it'll be easier to squash them later.

app/assets/javascripts/shared/flash_message.js
mqchau: I guess the longterm solution is to name those classes main-layout__flash-message-<error/notice/success> and have those classes inherit from Bootstrap classes in a *.scss file. I think this deserve a separate PR by itself.

mqchau: well there is a small translation:
flash notice -> alert-info
flash error -> alert-danger
flash success -> alert-success

So currently this logic happens in 2 places: set_flash_message.js and app/helpers/application_helpers.rb. I think this violates DRY and I think pushing this logic to *.scss file is the best. Since later on if Bootstrap decide to change class name again there's only one place to fix.

rmacklin: I would agree with you but just for the record the simpler way to be DRY is just to use bootstrap's names in your `flash` keys. They wouldn't change those names without a major version bump, and at the time you wanted to upgrade you could just find and replace your usages (or alias things to the old names). In any case, using your own classes rather than bootstrap's definitely gives the most flexibility so I'm not discouraging that, but I just wanted to point out that it's not the only way nor the simplest way to be DRY (especially considering bootstrap is unlikely to change those names, even in a new major version).

rmacklin: Just in case it's not clear, I was only replying to keep the discussion going, but I still think we don't need to change this in this PR.

test/flows/image_share_test.rb
rmacklin: Oh, you'd have to provide the `host` option to `share_image_url` as the test sever's host for that to work. It's probably fine to leave it like this.

app/models/image_share_form.rb
mqchau: Do you thin this fix is what you mean?

rmacklin: Yes that works. Since it's a partial it would be a bit cleaner to use local variables instead of instance variables, which would also eliminate the `@image_share_form || ImageShareForm.new` since the proper value would be passed in each case the partial is rendered. And you can replace `@image ? @image.link : ''` with `image.try(:link)`.

mqchau:
Hi Richard, in the last 6 commits I have refactor the image share form partial to use local variables only. I used image.try(:link).to_s because image.try(:link) return nil and image_tag will raise error when it's nil. Empty string seems to work though.

rmacklin:
> image_tag will raise error when it's nil

Oh interesting, that's a new change in Rails 5 (`nil` works in Rails 4). `to_s` seems fine. The changes look good.

mqchau:
Ok I'll rebase/rearrange things then.

mqchau:
So do you think this PR is good to merge then?

.rubocop.yml
```diff
@@ -12,6 +12,9 @@ Metrics/LineLength:
 ClassLength:
   Max: 500
 
+MethodLength:
+  Max: 50
```
rmacklin: 50 is super large. Can you reduce this?

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,48 @@
+import $ from 'jquery';
+import setFlashMessage from '../shared/set_flash_message';
+
+class ShareModal {
+  constructor(modalSelector){
+    this.$modal = $(modalSelector);
+    this.$form = this.$modal.find('form');
+    this.$form.$emailAddress = '';
```
rmacklin: We shouldn't tack this arbitrary attribute onto the form jQuery element. We can just use another instance property, e.g. `this.mostRecentRecipient` (it should _not_ be prefixed by `$` since it is not a jQuery object)

rmacklin:
From your commit message on 9811e8c8fb2ea3acb346a37a630b520affb02edb:
> Can't require SENDGRID_USERNAME and SENDGRID_PASSWORD in app.json
because it will break deployment

That shouldn't be the case. If you specify those as required in app.json, they will be inherited from your staging application. So as long as your staging app has the Sendgrid addon, it should work.

rmacklin:
I did a quick review and left 3 new comments. We can merge this after those are addressed.

mqchau:
The SENDGRID_USERNAME and SENDGRID_PASSWORD requirement breaks the app deployment when I created a new PR -> new app on Heroku from scratch. Otherwise if it updates a running app things will be OK. I don't think this behavior is desirable so I took them out.

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,48 @@
+import $ from 'jquery';
+import setFlashMessage from '../shared/set_flash_message';
+
+class ShareModal {
+  constructor(modalSelector){
+    this.$modal = $(modalSelector);
+    this.$form = this.$modal.find('form');
+    this.$form.mostRecentRecipient = '';
```
rmacklin: Again, don't tack this onto the jQuery element. Make it a property of the ShareModal instances: `this.mostRecentRecipient`

PR #43, opened 2016-09-08
Rename flash message class

mqchau:
Patch #6 

There's a translation between the name of flash message type and bootstrap alert classes. Now the translation logic is implemented in both controller and front end JS. To follow DRY principle, I move the logic into the sass template and use the alert message class such that it will follow the name of flash message type


app/assets/stylesheets/_main-layout.scss
```diff
@@ -8,3 +8,18 @@
   flex: 1;
   padding-top: 70px;
 }
+
+.main-layout__flash-message-error {
```
aaronpanch: May I suggest creating a totally new component here -- I don't see how the flash message has to be a part of the `main-layout` component.  If it has to do with positioning, I suggest making a `main-layout__flash-message` container subcomponent in main-layout, but still have the aliasing in it's own component.

app/views/layouts/_flash_message.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="main-layout__flash-message" id="flash_messages">
   <% flash.each do |key, value| %>
     <% unless value.empty? %>
-      <div id="<%= key %>" class="alert <%= flash_to_alert_class(key) %>"><%= value %></div>
+      <div class="main-layout__flash-message-<%= key %>"><%= value %></div>
```
aaronpanch: Interpolating with most of the content determined by ruby can get confusing.  Maybe it's better to use `content_tag` here or to extract this to a helper?  What are your thoughts?

app/assets/stylesheets/_main-layout.scss
mqchau: I was thinking of having flash-message as its own component too. Now u suggest it I'll do it.

app/assets/stylesheets/_flash-message.scss
```diff
@@ -0,0 +1,14 @@
+.flash-message__error {
```
aaronpanch: This is almost perfect -- next, we should convert each of these sub-components into modifiers.  So, I'm thinking something like this:

```
.flash-message { // the top-level component
  @extend .alert;

  &--error {
    @extend .alert-danger;
  }

  // and so on
}

File: app/assets/stylesheets/application.scss [hunk 4]
```diff
@@ -6,3 +6,4 @@ $card-spacer-x: 0.5rem;
 @import "navbar";
 @import "image";
 @import "main-layout";
+@import "flash-message";
```
aaronpanch: Alphabetize

app/views/layouts/_flash_message.html.erb
```diff
@@ -1,7 +1,8 @@
-<div class="main-layout__flash-message" id="flash_messages">
+<div class="flash-message" id="flash_messages">
   <% flash.each do |key, value| %>
     <% unless value.empty? %>
-      <div id="<%= key %>" class="alert <%= flash_to_alert_class(key) %>"><%= value %></div>
+      <%= content_tag :div, value, class: 'flash-message__' + key %>
```
aaronpanch: This looks better than the ERB interpolation, but I'm curious why you don't want to use the helper to translate from flash keys to CSS classes?

PR #44, opened 2016-09-09
New flow test

### LINKED ISSUE #37: Refactor flow tests to use page objects
Reference material:
 - https://github.com/appfolio/ae_page_objects

The starting skeleton is on the `flowTestsWithPageObjects` branch.

mqchau:
Closes #37


test/page_objects/images/elements/image_box.rb
```diff
@@ -0,0 +1,31 @@
...
+    class ImageBox < AePageObjects::Element
+      def link
+        node.find('img')[:src]
+      end
+
+      def tags
+        node.all('.js-image-tags').map(&:text)
+      end
+
+      def dismiss_destroy!
```
rmacklin: `!` methods have a special meaning in our page object conventions: only end a method in `!` if it returns a new page.

test/page_objects/images/elements/image_box.rb
```diff
@@ -0,0 +1,31 @@
...
+        parent.node.dismiss_confirm { node.click_link('Destroy') }
+      end
+
+      def accept_destroy!
+        parent.node.accept_confirm { node.click_link('Destroy') }
+        window.change_to(IndexPage)
+      end
```
rmacklin: I'm going to preface my next several comments with this: Designing a good page object interface is the most important thing to do when writing page objects. It makes writing tests easier, reading tests easier, modifying tests easier, and reduces the possibility of writing flaky or invalid tests.

rmacklin: With that said, this page object's interface is confusing. `dismiss_destroy` really performs an action corresponding to "`try_to_destroy_and_dismiss_confirmation`", so the naming isn't clear. Similarly, `accept_destroy` is more like `destroy_and_confirm`. Additionally, neither of these methods provide a way for tests to access the confirmation dialog itself to ask questions about it (for example, what text is displayed in the dialog?) prior to either dismissing or confirming it.

Here is a better interface:
```ruby
def destroy
  node.click_on('Destroy')
  yield node.session.driver.browser.switch_to.alert
end

def destroy_and_confirm!
  destroy(&:accept)
  window.change_to(IndexPage)
end
```
The first method simply clicks the _Destroy_ button and yields the confirmation dialog to the caller, so (unlike the current methods) it's completely flexible for however the tests want to use it. For example:
```ruby
image_show_page.destroy do |confirm_dialog|
  assert_equal 'Are you sure?', confirm_dialog.text
  confirm_dialog.dismiss
end
```

The second method provides a shortcut for the common case of "just destroy this image and accept the confirmation", which will redirect the user to the `IndexPage`.

rmacklin: Also, this interface was provided and used in the `flowTestsWithPageObjects` branch which is listed as the starting skeleton for issue #37. It seems you abandoned that branch and started from scratch, which is not desirable.

test/page_objects/images/elements/share_image_modal.rb
```diff
@@ -0,0 +1,24 @@
+module PageObjects
+  module Images
+    class ShareImageModal < AePageObjects::Element
+      form_for :image_share_form, locator: '#new_image_share_form' do
+        element :email_address, locator: '#image_share_form_email_address'
+        element :message, locator: '#image_share_form_message'
+      end
+
+      element :preview_image, locator: '.js-image-share-preview'
+      element :error_message, locator: '.form-control-feedback'
```
rmacklin: This is unnecessary. The `ElementInlineErrorMessage` extension provides an `error_message` method for use in all form elements.

test/page_objects/images/elements/image_box.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      def click_share!
+        node.click_on('Share')
+      end
```
rmacklin: The `click_share!` method combined with the static/always available `share_image_modal` element ([here](https://github.com/AppFolioOnboarding/image-sharer-mqchau/blob/943e35f8f194f516142aaf89eaba77aae9ec8b02/test/page_objects/images/pages/index_page.rb#L6) and [here](https://github.com/AppFolioOnboarding/image-sharer-mqchau/blob/943e35f8f194f516142aaf89eaba77aae9ec8b02/test/page_objects/images/pages/show_page.rb#L7)) is another problematic interface. By declaring `share_image_modal` as a "static" element on those pages via:
```ruby
class ShowPage
  element :share_image_modal, is: ShareImageModal
end
```
you are allowing access to a `share_image_modal` element even when such an element cannot be interacted with. This makes it very easy to write invalid tests as well as flaky tests.

A much better approach is to provide a dynamically instantiated element for the modal that can _only_ be accessed when it can legitimately be interacted with. This can be accomplished by constructing a dynamic element inside a method, as demonstrated in the ae_page_objects readme example for ["Creating elements on the fly"](https://github.com/appfolio/ae_page_objects/blob/013d705591e20b9f91601ebbfc8d5577ac32d714/README.md#creating-elements-on-the-fly).

Thus, we should provide an interface like:
```ruby
def open_share_modal
  node.click_on('Share')
  share_modal = document.element(locator: '#share_modal', is: ShareModal)
  share_modal.wait_until_visible
  share_modal
end

def share(&block)
  share_modal = open_share_modal
  block.call(share_modal)
  share_modal.wait_until_hidden
end
```

test/page_objects/images/elements/share_image_modal.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      def share!(email_address:, message: nil)
+        self.email_address.set email_address
+        self.message.set message if message.present?
+
+        node.click_on('Send email')
+      end
+
+      def close!
```
rmacklin: As mentioned [above](https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/44#discussion_r78437444), neither this nor the `share!` method should end in a `!`

test/page_objects/images/pages/edit_page.rb
```diff
@@ -0,0 +1,23 @@
...
+      end
+
+      def edit_tag_list!(tag_list: nil)
+        self.tag_list.set tag_list if tag_list.present?
+
+        node.click_button('Save changes')
+        window.change_to(ShowPage)
+      end
+
+      def link?(link)
```
rmacklin: This method name is confusing. From the implementation, it looks like the method is really more like `displaying_image?`

test/page_objects/images/pages/index_page.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+
+      def showing_image?(link:, tag_list: nil)
+        images.any? do |image|
+          result = (image.link == link)
+          tag_list.present? ? (result && image.tags == tag_list) : result
+        end
+      end
+
+      def find_image_by_link(link)
```
rmacklin: This method shouldn't exist in the page object. The images collection implements the `Enumerable` interface, which provides the flexibility for `find` and all other `Enumerable` methods to be called directly on the collection in the tests that need them (without needing a corresponding page object method for each `Enumerable` method).

test/page_objects/images/pages/edit_page.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Images
+    class EditPage < AePageObjects::Document
+      path :edit_image
+      path :image
+
+      form_for :image do
+        element :tag_list
+      end
+
+      def edit_tag_list!(tag_list: nil)
```
rmacklin: This is another method that isn't well named. Page objects methods should be named corresponding to actions the user would take, in the terminology the user would think of.

In this case, the user is on the page to edit an image. They got there by clicking an "Edit" button (not an "Edit tag list" or "Edit tags") button. The button to save their updates to the image is labeled "Save changes", so `save_changes!` would be a better name for this method.

test/page_objects/images/pages/edit_page.rb
```diff
@@ -0,0 +1,23 @@
...
+        self.tag_list.set tag_list if tag_list.present?
+
+        node.click_button('Save changes')
+        window.change_to(ShowPage)
```
rmacklin: `ShowPage` is not the only place you can go when you submit this form. If the user enters an blank string (`edit_page.tag_list.set(' ')`) and then submits the form (`edit_page.save_changes!`), they'll remain on the edit page. Your flow tests need to include this case which means your page object needs to support it.

rmacklin: Also if you get stuck on this change, you should discuss it with Peter.

test/page_objects/images/pages/show_page.rb
```diff
@@ -0,0 +1,15 @@
+module PageObjects
+  module Images
+    class ShowPage < PageObjects::Document
+      path :image
+
+      element :image, locator: '.js-image-box', is: ImageBox
```
rmacklin: You should delegate methods directly to the `image` to improve the page object interface of the show page. Lines like:
```ruby
images_edit_page = image_show_page.image.edit!
```
would be clearer/cleaner/simpler as:
```ruby
images_edit_page = image_show_page.edit!
```
and we can use `delegate` to provide this improved interface. Ask Peter if you need help with this.

test/page_objects/images/pages/index_page.rb
```diff
@@ -0,0 +1,34 @@
+module PageObjects
+  module Images
+    class IndexPage < PageObjects::Document
+      path :images
+
+      element :share_image_modal, is: ShareImageModal
+      collection :images,
+                 locator: '.main-layout__content',
```
rmacklin: You should use a specific locator that will always be the container for the images collection, rather than a style class that could change/be removed (which shouldn't break our flow tests).

test/page_objects/images/pages/index_page.rb
```diff
@@ -0,0 +1,34 @@
...
+                 item_locator: [
+                   :xpath,
+                   # this XPath code will select all .js-image-box inside different col-md-*
+                   "descendant::div[contains(concat(' ', normalize-space(@class), ' '), ' js-image-box ')]"
```
rmacklin: It's not clear why you're using xpath to select images. In general, we want to avoid xpath in our page objects as much as possible.

rmacklin:
The `pages/` and `elements/` folders split related page object classes on the file system unnecessarily. This should be avoided. I suspect this directory structure may have only been created so that you could require the elements before the pages, on [these lines](https://github.com/AppFolioOnboarding/image-sharer-mqchau/blob/943e35f8f194f516142aaf89eaba77aae9ec8b02/test/flow_test_helper.rb#L14-L15). If so, the better solution is just to require the elements that each page uses at the top of the file for each page.

test/page_objects/images/elements/image_box.rb
rmacklin: The `share` method should be used the majority of the time. The `open_share_modal` method is only necessary for testing the case where you attempt to share a deleted image, because that case redirects the user to a refreshed index page, which requires calling `AePageObjects.browser.current_window.change_to(PageObjects::Images::IndexPage)` in the test, like:
```ruby
images_index_page = PageObjects::Images::IndexPage.visit

share_modal = images_index_page.images.first.open_share_modal

Image.last.destroy!

share_modal.email.set 'valid@email.com'
share_modal.message.set 'Look at this image'
share_modal.share_image

images_index_page =
  AePageObjects.browser.current_window.change_to PageObjects::Images::IndexPage

assert_equal "This image has been deleted",
             images_index_page.flash_message(:failure)
```

PR #45, opened 2016-09-09
Qc authenticate

### LINKED ISSUE #11: Authentication
#### As an application hoster I want to add user authentication to my application.

- [ ] Users can sign up.
- [ ] Users can log in after they've signed up.
- [ ] Users can log out after logging in.
- [ ] Users can choose to be remembered after logging in.
- [ ] Multiple browsers can concurrently hold permanent cookies.
- [ ] Logged in users should not be able to register, nor log in again.

__Note__: Do not use Devise for this task.

Discussion topics:
- [ ] If we wanted to add more advanced features such as email activation and
  "forgot my password", should we continue to "roll our own" authentication?

Reference material:
- https://www.railstutorial.org/book/log_in_log_out

Further reading:
- https://github.com/plataformatec/devise

Dependencies:
- Delete Image

mqchau:
Closes #11


Gemfile
```diff
@@ -21,7 +21,7 @@ gem 'jbuilder', '~> 2.5'
 # Use Redis adapter to run Action Cable in production
```
bboe: > bcrypt to support user model

Can this be worded better?

test/models/user_test.rb
```diff
@@ -0,0 +1,51 @@
+require 'test_helper'
+
+class UserTest < ActiveSupport::TestCase
+  test 'invalid user with no name' do
+    user = User.new(name: '', password: 'abc')
+    refute_predicate user, :valid?
```
bboe: What else can you assert when validation fails?

Gemfile
mqchau: I'll rename this when there's change to rebase the entire history

test/models/user_test.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  test 'invalid user with duplicated username' do
+    User.create!(name: 'abc', password: 'abc')
+    user = User.new(name: 'abc', password: 'cde')
+    assert_predicate user, :invalid?
+    assert_includes user.errors.full_messages, 'Name has already been taken'
+  end
+
+  test 'invalid user with mismatch password confirmation' do
```
bboe: Great test!

test/models/user_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    user = User.new(name: 'abc', password: 'abc', password_confirmation: 'ab')
+    assert_predicate user, :invalid?
+    assert_includes user.errors.full_messages, "Password confirmation doesn't match Password"
+  end
+
+  test 'invalid user retrieval with wrong password' do
+    user = User.new(name: 'abc', password: 'abc')
+    refute_operator user, :authenticate, 'ab'
```
bboe: Also a great test.

test/models/user_test.rb
```diff
@@ -2,50 +2,54 @@
 
 class UserTest < ActiveSupport::TestCase
   test 'invalid user with no name' do
-    user = User.new(name: '', password: 'abc')
+    user = User.create(name: '', password: 'abc')
```
bboe: Do these need to be persisted to the database?

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,21 @@
+class UsersController < ApplicationController
+  def new
+    @user = User.new
+  end
+
+  def create
+    @user = User.new(user_params)
+    if @user.save
+      redirect_to root_path
```
bboe: Should there be a message what the account is created?

app/views/layouts/_user.html.erb
```diff
@@ -0,0 +1,3 @@
+  <li class="nav-item <%= 'active' if current_page?(new_user_path) || current_page?(users_path) %>">
```
bboe: Should `current_page?` be adapted to take more than one argument?

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,51 @@
+require 'test_helper'
+
+class UsersControllerTest < ActionDispatch::IntegrationTest
+  test 'should get signup' do
+    get new_user_path
+
+    assert_response :success
```
bboe: Please use `:ok` instead of success. If you're using `:success` elsewhere, please make a pull request to convert `:success` to a more precise status code.

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  test 'should not signup user with invalid info' do
+    assert_no_difference('User.count') do
+      post users_path, params: {
+        user: { name: 'admin', password: 'pa', password_confirmation: 'pac' }
+      }
+    end
+
+    assert_response :unprocessable_entity
+    assert_equal 'Can not create new account', flash[:error]
```
bboe: Should there also be an error message on a form field?

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  test 'should not signup users with duplicate name' do
```
bboe: This test isn't really necessary as your model already tests it. The controller test should simply include one failure case to ensure that failures, in general, work they way they should.

test/flows/user_flow_test.rb
```diff
@@ -0,0 +1,19 @@
+require 'flow_test_helper'
+
+class UserFlowTest < FlowTestCase
+  test 'sign up' do
```
bboe: This test is great for now, but make sure this is merely incorporated as part of more defined flow somewhere. Otherwise it looks pretty similar to the controller test for account sign up, and does not add much value.

test/flows/user_flow_test.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    sign_up_page = sign_up_page
+      .create_user!(name: 'admin', password: 'password', password_confirmation: 'passwod')
+      .as_a(PageObjects::Users::SignUpPage)
+
+    assert_match 'Can not create new account', sign_up_page.flash_message(:error)
+
+    sign_up_page.password.set('password')
+    sign_up_page.password_confirmation.set('password')
+    images_index_page = sign_up_page.create_user!
```
bboe: Could you pass in the password and its confirmation to `create_user` for simplicity?

app/assets/stylesheets/_navbar.scss
```diff
@@ -4,3 +4,15 @@
   font-size: 120%;
```
bboe: > work on sign in

What does this mean?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    if user.nil? || !user.authenticate(session_params[:password])
+      flash.now[:error] = 'Invalid name/password combination.'
+      render :new, status: :unauthorized
+    else
+      sign_in user
+      redirect_to root_path
+    end
+  end
+
+  def session_params
```
bboe: `params.require` isn't necessary if you aren't using `update_attributes` or similar (passing a hash). In `create` you're extracting each field individually which is perfect.

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,19 @@
+class SessionsController < ApplicationController
+  def new
+  end
+
+  def create
+    user = User.find_by(name: session_params[:name])
+    if user.nil? || !user.authenticate(session_params[:password])
+      flash.now[:error] = 'Invalid name/password combination.'
```
bboe: ⭐ awesome for not differentiating between invalid user and invalid password. No reason to give malicious users extra information.

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,18 @@
+module SessionsHelper
+  def sign_in(user)
+    session[:user_id] = user.id
+  end
+
+  def current_user
+    return unless (user_id = session[:user_id])
```
bboe: WHere does `user_id` come from?

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+    @current_user ||= User.find_by(id: user_id)
+  end
+
+  def signed_in?
+    !current_user.nil?
```
bboe: Do you like the use of `!` combined with `nil?`?

app/helpers/sessions_helper.rb
bboe: Also, what does `user_id` represent?

app/views/layouts/_user.html.erb
```diff
@@ -1,3 +1,12 @@
+<% if not signed_in? %>
```
bboe: Do you think `unless` would be better suited here?

bboe: Or better, structuring this such that the first section is for signed in, and the second section is for the `else`.

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,49 @@
+require 'test_helper'
+
+class SessionsControllerTest < ActionDispatch::IntegrationTest
+  test 'should get signin' do
+    get new_session_path
+
+    assert_response :success
```
bboe: :success -> :ok

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,49 @@
...
+    assert_select 'form#new_session input' do
+      assert_select '[name=?]', 'session[name]'
+      assert_select '[name=?]', 'session[password]'
+      assert_select '[type=?]', 'submit'
+    end
+    assert_select 'form#new_session[action=?]', sessions_path
+    assert_select 'form#new_session[method=?]', 'post'
+  end
+
+  test 'should signin old user' do
```
bboe: Is `old` the best word to use? How about `existing`?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  test 'should not signin user with wrong password' do
+    User.create!(name: 'admin', password: 'password')
+    log_in_as('admin', password: 'passwod')
+
+    assert_response :unauthorized
+    assert_equal 'Invalid name/password combination.', flash[:error]
+  end
+
+  # Log in as a particular user.
```
bboe: Should there be a `private` section?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,49 @@
...
+  def log_in_as(username, password: 'password', remember_me: '0')
```
bboe: Does `remember_me` exist at this point?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,49 @@
...
+    post sessions_path, params: {
+      session: { name: username, password: password, remember_me: remember_me }
+    }
+  end
+
+  def assert_logged_in(username)
```
bboe: These assertions make an additional request (well mimic it). Recall that with controller tests we should avoid issuing additional requests so that we are only testing the one. How else can you assert that one is logged in without making a request?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,49 @@
...
+    get root_path
+    assert_select '.navbar__username', username
+  end
+
+  def assert_not_logged_in
+    get root_path
```
bboe: Similar issue as https://github.com/AppFolioOnboarding/image-sharer-mqchau/pull/45/commits/1a373deb8ca7cb371bcbcf363f202a023104d657#r78431958

test/flows/user_flow_test.rb
```diff
@@ -1,7 +1,7 @@
 require 'flow_test_helper'
 
 class UserFlowTest < FlowTestCase
-  test 'sign up' do
+  test 'sign up, sign in' do
```
bboe: ⭐ Glad to see that you are extending this test.

test/page_objects/images/pages/index_page.rb
```diff
@@ -34,6 +36,15 @@ def sign_up!
         node.click_on('Sign up')
         window.change_to(PageObjects::Users::SignUpPage)
       end
+
+      def sign_in!
```
bboe: I wonder if these would be better suited as methods on all page objects since they are part of the layout. I am not actually 100% sure how we do this in property.

test/models/user_test.rb
mqchau: I tried User.new then user.valid?, user.errors is empty after that.

app/controllers/users_controller.rb
```diff
@@ -6,6 +6,7 @@ def new
   def create
     @user = User.new(user_params)
     if @user.save
+      sign_in(@user)
```
bboe: This results in a side-effect. It should be tested for in the user create test.

app/controllers/sessions_controller.rb
```diff
@@ -13,6 +13,11 @@ def create
     end
```
bboe: > work on sign out

What does this mean?

app/views/layouts/_user.html.erb
```diff
@@ -9,4 +9,7 @@
   <li class="nav-item navbar__container">
     <p class="navbar__username-area">Hello <span class="navbar__username"><%= current_user.name %></span></p>
   </li>
+  <li class="nav-item">
+    <%= link_to 'Sign out', destroy_session_path, method: :delete, class: 'nav-link' %>
```
bboe: `link_to` should really only be used for links because `a` tags can be opened via a few methods (cmd+click, middle click) which result in unexpected behavior. Buttons can only be clicked on (by default) and thus are preferred.

bboe: If you use `link_to` elsewhere where it's not just loading a show-page, it should be replaced with `button_to`.

config/routes.rb
```diff
@@ -9,4 +9,5 @@
 
   resources :users, only: [:create, :new]
   resources :sessions, only: [:create, :new]
+  delete '/sessions', to: 'sessions#destroy', as: 'destroy_session'
```
bboe: Why not add this to the `resources` definition?

test/controllers/sessions_controller_test.rb
```diff
@@ -30,6 +30,23 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'Invalid name/password combination.', flash[:error]
   end
 
+  test 'should signout user' do
```
bboe: Does this actually work the way you intend it to?

app/views/layouts/_user.html.erb
mqchau: Vanilla current_page? takes only 1 argument, according to this: http://apidock.com/rails/ActionView/Helpers/UrlHelper/current_page%3F

Do you mean to wrap it in a helper method?

test/flows/user_flow_test.rb
```diff
@@ -14,6 +14,31 @@ class UserFlowTest < FlowTestCase
...
+    sign_in_page.name.set('admin')
+    sign_in_page.password.set('password')
+
+    images_index_page = sign_in_page.sign_in_user!
+
+    assert images_index_page.signed_in?('admin')
+
+    images_index_page = images_index_page.sign_out!
+
+    refute images_index_page.signed_in?('admin')
```
bboe: ⭐ great test

app/controllers/sessions_controller.rb
```diff
@@ -21,4 +23,10 @@ def destroy
   def session_params
     params.require(:session).permit(:name, :password)
   end
+
+  private
+
+  def check_sign_in
+    redirect_to root_path if signed_in?
```
bboe: Should there be a message to the user as to why they were redirected? Also, this method seems like it does more than just check if the user is signed in. Can the name be more precise?

test/controllers/sessions_controller_test.rb
```diff
@@ -47,6 +47,49 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to root_path
   end
 
+  test 'signed in user should not get sign up page' do
+    User.create!(name: 'admin', password: 'password')
+    log_in_as('admin')
+
+    get new_user_path
```
bboe: Again there is more than one "request" in these tests. How can you "sign in" a user without issuing a request for it?

app/controllers/sessions_controller.rb
```diff
@@ -5,12 +5,13 @@ def new
   end
```
bboe: > remember_me

I'm being pedantic now, but please ensure your commit messages are proper sentences.

test/controllers/sessions_controller_test.rb
```diff
@@ -47,6 +47,49 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
...
+  test 'signed in user should not get sign in page' do
+    User.create!(name: 'admin', password: 'password')
+    log_in_as('admin')
+
+    get new_session_path
+
+    assert_redirected_to root_path
+  end
+
+  test 'signed in user should not sign in again' do
```
bboe: I really like that you've handled these cases and tested for it. Let's just try to make the tests a bit more robust by not requiring multiple "requests". It would also be great to have assertions for "you are already signed in" type messages.

app/controllers/sessions_controller.rb
```diff
@@ -21,12 +22,17 @@ def destroy
...
 
   private
 
   def check_sign_in
     redirect_to root_path if signed_in?
   end
+
+  def authenticate
+    user = User.find_by(name: session_params[:name])
```
bboe: While not a major issue. Can we save a DB query if name doesn't exist in the session?

app/helpers/sessions_helper.rb
```diff
@@ -4,14 +4,25 @@ def sign_in(user)
   end
 
   def current_user
-    return unless (user_id = session[:user_id])
-    @current_user ||= User.find_by(id: user_id)
+    if (user_id = session[:user_id])
+      @current_user ||= User.find_by(id: user_id)
+    elsif (user_id = cookies.signed[:user_id])
+      user = User.find_by(id: user_id)
+      if user && user.authenticated?(cookies[:remember_token])
```
bboe: Can a user be concurrently and persistently signed in on two browsers?

bboe:
Made my first pass through. Please go ahead and reword and rebase where necessary. I'll make a complete second pass through.

test/models/user_test.rb
bboe: Oh interesting. Maybe a comment should be added then that errors is only actually added when attempting to create. I wonder if there is a method you could call that triggers the bcrypt gem validation that doesn't require persisting to the DB.

app/views/layouts/_user.html.erb
bboe: Yeah, sorry, I mistakenly assumed this was something you wrote. Since you use this `||` pattern in a few places it might be nice to make your own helper for it.

